Issue #2, opened 2024-04-14
Add config validation


theanmolsharma:
Install and use `class-validator` to validate config in `configuration.ts`

Nesopie:
Hey! Can I work on this?

theanmolsharma:
of course, I'm assigning it to you. We can use the issue for discussion if you have any questions. We also have a weekly sync call [Discord](https://discord.com/events/1104831812720656540/1209759462768775168) on Tuesday which you can join.

Nesopie:
thanks! I'll be sure to join it :)

PR #6, opened 2024-04-15
feat: add db config validation

Nesopie:
PR for Issue #2 
Very new to nest.js but for some reason the app continues to run even when I'm throwing an error?


theanmolsharma:
Few thoughts:
- We should perform validation inside `configuration.ts`. This will prevent the app from starting if the validation fails.
- Let's rename the file that contains the validation class `DbConfig` to `src/configuration.model.ts`. The name you choose is too specific. As we add more config values, we may want to add different classes to the same file.
- I don't think you need to define a constructor. The class should just be a blueprint for validation. We are not planning to create objects for the class.
- I think you can just call `validate<DbConfig>(config, DbConfig);` instead of first instantiating an object, then performing the validation.

Nesopie:
Hey @theanmolsharma, turns out you can't validate [plain objects without instantiating a class](https://github.com/typestack/class-validator?tab=readme-ov-file#validating-plain-objects) so I had to make use of `class-transformer` which I needed anyways, since I'm validating nested objects.

theanmolsharma:
@Nesopie are you still on this?

Nesopie:
@theanmolsharma It's ready for review. Is there anything that you feel is missing / not right?

src/configuration.model.ts
theanmolsharma, line 1: let's get rid of custom error messages, I think the error returned by class-validator is pretty self-explanatory and you are not adding any extra useful info with custom messages.

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+    Min,
+    MinLength,
+    ValidateNested,
+} from 'class-validator';
+import { Type } from 'class-transformer';
+
+class DbConfig {
+    @MinLength(1, {
+        message: 'Host length should at least be 1',
+    })
```
theanmolsharma, line 16: instead of `@MinLength(1)`, let's use `@IsNotEmpty()`

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+    @IsString({
+        message: 'Host should be a string',
+    })
+    host: string;
+
+    @IsNumber(
+        {
+            allowNaN: false,
```
theanmolsharma, line 24: i think the default for `allowNaN` is `false`, so let's get rid of this

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+class DbConfig {
+    @MinLength(1, {
+        message: 'Host length should at least be 1',
+    })
+    @IsString({
+        message: 'Host should be a string',
+    })
+    host: string;
+
+    @IsNumber(
```
theanmolsharma, line 22: let's use `IsInteger` instead, because port cannot be floating point number

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+        message: 'Port should at least be 0',
+    })
+    @Max(65535, {
+        message: 'Port should at most be 65535',
+    })
+    port: number;
+
+    @MinLength(1, {
+        message: 'Username length should at least be 1',
+    })
```
theanmolsharma, line 40: let's do `IsString` and `IsNotEmpty`

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+    username: string;
+
+    password: string;
```
theanmolsharma, line 43: i think you forgot to add validation for this property

src/configuration.model.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    @MinLength(1, {
+        message: 'Database name length should at least be 1',
+    })
```
theanmolsharma, line 47: same comment here, use `IsString` and `IsNotEmpty`

src/configuration.ts
```diff
@@ -2,6 +2,9 @@ import { join } from 'path';
 import { readFileSync } from 'fs';
 import { load } from 'js-yaml';
 import { camelToSnakeCase } from '@/common/common';
+import { Config } from './configuration.model';
```
theanmolsharma, line 5: Shouldn't this be `import { Config } from '@/configuration.model';`?

src/configuration.ts
```diff
@@ -40,5 +43,13 @@ export default () => {
...
     >;
     mergeEnvVariablesRecursive(config);
+
+    validate(plainToClass(Config, config)).then((errors) => {
+        for (const error of errors) {
+            console.error(error.toString());
+        }
+
+        if (errors.length > 0) throw new Error('Invalid config');
+    });
```
theanmolsharma, line 53: I'm pretty sure that you should be able to just do something like
```ts
    const validatedConfig = plainToInstance(Config, config, {
        enableImplicitConversion: true,
    });
    const errors = validateSync(validatedConfig as object, {
        skipMissingProperties: false,
    });

    if (errors.length > 0) {
        throw new Error(errors.toString());
    }

    return validatedConfig;
```

I don't like chained callback (`.then()`)

theanmolsharma (changes requested):
Apart from addressing comments, I'd appreciate if you could do the following:
1. squash commits into just 2 commits. One for adding the dependencies and one for adding validation logic.
2. sign all your commits.

Awesome work!! Thanks for putting in the time and effort for contributing.

src/configuration.model.ts
Nesopie: I've seen passwords that are empty so I didn't add anything, I'll just add the same validation as username for now though.

PR #14, opened 2024-06-04
Improved Common Lib Unit Test Readability

aruokhai:
### Objective

The objective of this change is to improve the readability of the _common_ lib module unit test, which is necessary for effective testing, given that a peculiar function `extractPubKey` in the _common_ lib is quite crucial for the correctness of the Project, 

### Problem

The **common** lib module  is not  utilising  behavioural test patterns which makes isolating a specific branch of the code under test quite difficult. The problem is quite critical for functions with extreme number of branch paths such as  `extractPubKey` .

### Changes

-  Added a test fixtures file  in the _common_ lib module, which contains the necessary data need for testing all common lib functions.
- Added behavioural unit test pattern in the common lib module test file and separated a monolithic test case to branch based test cases.

### Scope Of Change

This is a non critical change because it only affects a unit test file.

### DIFF
--- a/src/common/common.spec.ts
+++ b/src/common/common.spec.ts
@@ -1,185 +1,94 @@
 import { extractPubKeyFromScript } from '@/common/common';
+import {
+    OutputScriptFixture,
+    p2pkh_invalid_fixtures,
+    p2pkh_valid_fixtures,
+    p2sh_p2wpkh_invalid_fixtures,
+    p2sh_p2wpkh_valid_fixtures,
+    p2tr_invalid_fixtures,
+    p2tr_valid_fixtures,
+    p2wpkh_valid_fixtures,
+} from '@/common/test_fixtures';
 
 describe('Common', () => {
-    it.each([
-        {
-            description: 'P2PKH',
-            scriptPubKey: '76a91419c2f3ae0ca3b642bd3e49598b8da89f50c1416188ac',
-            scriptSig:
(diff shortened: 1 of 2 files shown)


theanmolsharma (changes requested):
I don't see any utility in splitting the fixtures by type. I'm all for moving the fixtures to a dedicated file called `fixtures.ts` but the way you've done it adds more duplicate code to the tests.

Action items:
- just add one array with all fixtures in a file called `fixtures.ts`
- revert splitting the tests into multiple `it` blocks

aruokhai:
> I don't see any utility in splitting the fixtures by type. I'm all for moving the fixtures to a dedicated file called `fixtures.ts` but the way you've done it adds more duplicate code to the tests.
> 
> Action items:
> 
> * just add one array with all fixtures in a file called `fixtures.ts`
> * revert splitting the tests into multiple `it` blocks

The basic idea of my pattern is to isolate and thoroughly test each code path. I would love it if more explanation is given why it is unnecessary.

theanmolsharma:
> > I don't see any utility in splitting the fixtures by type. I'm all for moving the fixtures to a dedicated file called `fixtures.ts` but the way you've done it adds more duplicate code to the tests.
> > Action items:
> > 
> > * just add one array with all fixtures in a file called `fixtures.ts`
> > * revert splitting the tests into multiple `it` blocks
> 
> The basic idea of my pattern is to isolate and thoroughly test each code path. I would love it if more explanation is given why it is unnecessary.

The isolation of each execution branch comes at the following cost:
- Code duplication
- Increased maintenance overhead

For this particular change, the cons outweigh the benefits.

PR #15, opened 2024-06-14
Support For BitcoinCore RPC Provider

### LINKED ISSUE #12: Block Data Provider using Bitcoin Core RPC
### Overview
A block data provider that connects to a Bitcoin Core Full Node and creates indexes.

### Constraints
- [`txindex`](https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch03.asciidoc#txindex) should not be required for the Bitcoin node

### RPCs that could be used to sync
- [`getblock`](https://developer.bitcoin.org/reference/rpc/getblock.html) with `verbosity = 2`
- [`getrawtransaction`](https://developer.bitcoin.org/reference/rpc/getrawtransaction.html) with `verbose = true` and `blockhash`

### Implementation Details
- Block Provider should emit [`IndexTransactionCommand`](https://github.com/Bitshala-Incubator/silent-pay-indexer/blob/33c988312aa24fd66958e836fe9fef7a78e5b702/src/commands/impl/index-transaction.command.ts#L14C14-L14C37) with the necessary data
- Add another command called `IndexBlockCommand` and it's handler.
- This block provider should use `IndexBlockCommand`.

aruokhai:
### Objective
The objective this change is to add support for Bitcoincore RPC Api.

### Problem 
The silent indexer relies on the consumption of external Api's in order to be fed Bitcoin Block data needed for indexing silent payments

### Changes
- Add logic to fetch block data from a **bitcoincore** rpc endoint.  
- Added Unit test for the provided logic.
- Modified  **config module** to add support for bitcoincore RPC config details.
- Modified **Block Provider module** to switch between Esplora and Bitcoincore depending on the provided config data. 

### Scope Of Change
This is a critical change because it affects the correctness of the silent-indexer.

closes #12

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,12 +16,12 @@
         "@nestjs/passport": "^10.0.0",
         "@nestjs/platform-express": "^10.3.7",
         "@nestjs/platform-socket.io": "^10.3.7",
-        "@nestjs/schedule": "3.0.0",
+        "@nestjs/schedule": "^3.0.0",
         "@nestjs/swagger": "^7.3.1",
         "@nestjs/typeorm": "^10.0.2",
         "@nestjs/websockets": "^10.3.7",
         "axios": "^1.7.2",
-        "bitcoin-core": "^4.2.0",
+        "currency.js": "^2.0.4",
         "js-yaml": "^4.1.0",
         "pg": "^8.11.5",
         "reflect-metadata": "^0.1.13",
@@ -770,7 +770,7 @@
       "version": "0.8.1",
       "resolved": "https://registry.npmjs.org/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz",
       "integrity": "sha512-IchNf6dN4tHoMFIn/7OE8LWZ19Y6q/67Bmf6vnGREv8RSbBVb9LPJxEcnwrcwX6ixSvaiGoomAUvu4YSxXrVgw==",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "@jridgewell/trace-mapping": "0.3.9"
       },
@@ -782,7 +782,7 @@
       "version": "0.3.9",
       "resolved": "https://registry.npmjs.org/@jridgewell/trace-mapping/-/trace-mapping-0.3.9.tgz",
       "integrity": "sha512-3Belt6tdc8bPgAtbcmdtNJlirVoTmEb5e2gC94PnkwEW9jI6CAHUeoG85tjWP5WquqfavoMtMwiG4P926ZKKuQ==",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "@jridgewell/resolve-uri": "^3.0.3",
         "@jridgewell/sourcemap-codec": "^1.4.10"
@@ -1392,7 +1392,7 @@
       "version": "3.1.2",
       "resolved": "https://registry.npmjs.org/@jridgewell/resolve-uri/-/resolve-uri-3.1.2.tgz",
       "integrity": "sha512-bRISgCIjP20/tbWSPWMEi54QVPRZExkuD9lJL+UIxUKtwVJA8wW1Trb1jMs1RFXo1CBTNZ/5hpC9QvmKWdopKw==",
-      "dev": true,
+      "devOptional": true,
       "engines": {
         "node": ">=6.0.0"
       }
@@ -1420,7 +1420,7 @@
       "version": "1.4.15",
       "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.4.15.tgz",
       "integrity": "sha512-eF2rxCRulEKXHTRiDrDy6erMYWqNw4LPdQ8UQA4huuxaQsVeRPFl2oM8oDGxMFhJUWZf9McpLtJasDDZb/Bpeg==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@jridgewell/trace-mapping": {
       "version": "0.3.25",
@@ -1987,25 +1987,25 @@
       "version": "1.0.11",
       "resolved": "https://registry.npmjs.org/@tsconfig/node10/-/node10-1.0.11.tgz",
       "integrity": "sha512-DcRjDCujK/kCk/cUe8Xz8ZSpm8mS3mNNpta+jGCA6USEDfktlNvm1+IuZ9eTcDbNk41BHwpHHeW+N1lKCz4zOw==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@tsconfig/node12": {
       "version": "1.0.11",
       "resolved": "https://registry.npmjs.org/@tsconfig/node12/-/node12-1.0.11.tgz",
       "integrity": "sha512-cqefuRsh12pWyGsIoBKJA9luFu3mRxCA+ORZvA4ktLSzIuCUtWVxGIuXigEwO5/ywWFMZ2QEGKWvkZG1zDMTag==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@tsconfig/node14": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/@tsconfig/node14/-/node14-1.0.3.tgz",
       "integrity": "sha512-ysT8mhdixWK6Hw3i1V2AeRqZ5WfXg1G43mqoYlM2nc6388Fq5jcXyr5mRsqViLx/GJYdoL0bfXD8nmF+Zn/Iow==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@tsconfig/node16": {
       "version": "1.0.4",
       "resolved": "https://registry.npmjs.org/@tsconfig/node16/-/node16-1.0.4.tgz",
       "integrity": "sha512-vxhUy4J8lyeyinH7Azl1pdd43GJhZH/tP2weN8TntQblOY+A0XbT8DJk1/oCPuOOyg/Ja757rG0CgHcWC8OfMA==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -2304,7 +2304,7 @@
       "version": "13.11.9",
       "resolved": "https://registry.npmjs.org/@types/validator/-/validator-13.11.9.tgz",
       "integrity": "sha512-FCTsikRozryfayPuiI46QzH3fnrOoctTjvOYZkho9BTFLCOZ2rgZJHMOVgCOfttjPJcgOx52EpkY0CMfy87MIw==",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/@types/yargs": {
       "version": "17.0.32",
@@ -2515,29 +2515,6 @@
       "integrity": "sha512-zuVdFrMJiuCDQUMCzQaD6KL28MjnqqN8XnAqiEq9PNm/hCPTSGfrXCOfwj1ow4LFb/tNymJPwsNbVePc1xFqrQ==",
       "dev": true
     },
-    "node_modules/@uphold/request-logger": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@uphold/request-logger/-/request-logger-2.0.0.tgz",
-      "integrity": "sha512-UvGS+v87C7VTtQDcFHDLfvfl1zaZaLSwSmAnV35Ne7CzAVvotmZqt9lAIoNpMpaoRpdjVIcnUDwPSeIeA//EoQ==",
-      "dependencies": {
-        "uuid": "^3.0.1"
-      },
-      "engines": {
-        "node": ">=4"
-      },
-      "peerDependencies": {
-        "request": ">=2.27.0"
-      }
-    },
-    "node_modules/@uphold/request-logger/node_modules/uuid": {
-      "version": "3.4.0",
-      "resolved": "https://registry.npmjs.org/uuid/-/uuid-3.4.0.tgz",
-      "integrity": "sha512-HjSDRw6gZE5JMggctHBcjVak08+KEVhSIiDzFnT9S9aegmp85S/bReBVTb4QTFaRNptJ9kuYaNhnbNEOkbKb/A==",
-      "deprecated": "Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.",
-      "bin": {
-        "uuid": "bin/uuid"
-      }
-    },
     "node_modules/@webassemblyjs/ast": {
       "version": "1.12.1",
       "resolved": "https://registry.npmjs.org/@webassemblyjs/ast/-/ast-1.12.1.tgz",
@@ -2712,7 +2689,7 @@
       "version": "8.11.3",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.11.3.tgz",
       "integrity": "sha512-Y9rRfJG5jcKOE0CLisYbojUjIrIEE7AGMzA/Sm4BslANhbS+cDMpgBdcPT91oJ7OuJ9hYJBx59RjbhxVnrF8Xg==",
-      "dev": true,
+      "devOptional": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -2742,7 +2719,7 @@
       "version": "8.3.2",
       "resolved": "https://registry.npmjs.org/acorn-walk/-/acorn-walk-8.3.2.tgz",
       "integrity": "sha512-cjkyv4OtNCIeqhHrfS81QWXoCBPExR/J62oyEqepVw8WaQeSqpW2uhuLPh1m9eWhDuOo/jUXVTlifvesOWp/4A==",
-      "dev": true,
+      "devOptional": true,
       "engines": {
         "node": ">=0.4.0"
       }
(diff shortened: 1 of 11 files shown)


theanmolsharma (changes requested):

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+export class BitcoinCoreProvider
+    extends BaseBlockDataProvider
+    implements OnApplicationBootstrap
+{
+    protected readonly logger = new Logger(BitcoinCoreProvider.name);
+    protected readonly operationStateKey = 'bitcoincore-operation-state';
+    private isSyncing = false;
+    private config: BitcoinCoreConfig;
+
+    public client: Client;
```
theanmolsharma, line 36: any reason why `client` is public?

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    public constructor(
+        private configService: ConfigService,
```
theanmolsharma, line 39:
```suggestion
        private readonly configService: ConfigService,
```

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+        }
+        this.isSyncing = false;
+    }
+
+    async getTipHeight(): Promise<number> {
+        try {
+            const body = {
+                method: 'getblockcount',
+                params: [],
+                ...this.createPartialRPCBody(),
```
theanmolsharma, line 112: If we have to call `this.createPartialRPCBody()` everytime before calling `this.request()` then it's better to get rid of this function and move it inside `request()`

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+            };
+            return await this.request(body);
```
theanmolsharma, line 114:
```suggestion
            return this.request(body);
```

theanmolsharma: * do this everywhere

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+            const body = {
+                method: 'getblockhash',
+                params: [height],
+                ...this.createPartialRPCBody(),
+            };
+            return await this.request(body);
+        } catch (error) {
+            this.logger.log(`Error fetching  block hash of height : ${height}`);
+            throw error;
+        }
```
theanmolsharma, line 132: Let's remove try-catch from here and move it inside `this.request()`. We'll probably end up moving it there once we implement retry logic so it's better to do it now.

theanmolsharma: Also, log the request body and response so that we can know what went wrong.

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    async getBlock(hash: string, verbosity: number): Promise<Block> {
```
theanmolsharma, line 135: mark these as private please. everything that is internal should be private.

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+        try {
+            const body = {
+                method: 'getblock',
+                params: [hash, verbosity],
+                ...this.createPartialRPCBody(),
+            };
+            return await this.request(body);
+        } catch (error) {
+            this.logger.log(`Error fetching block with block hash : ${hash}`);
```
theanmolsharma, line 144:
```suggestion
            this.logger.error(`Error fetching block with block hash : ${hash}`);
```
same suggestion everywhere, please use proper log levels

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+            blockHash,
+        };
+    }
+
+    private async parseTransactionInput(
+        txnInput: Input,
+    ): Promise<TransactionInput> {
+        let txid =
+            '0000000000000000000000000000000000000000000000000000000000000000';
+        let vout = 0;
```
theanmolsharma, line 214: IIRC the vout is set to `0xfffffffff` for coinbase txs

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    public async processBlock(height: number): Promise<Transaction[]> {
+        const parsedTransactionList: Transaction[] = [];
+        const blockHash = await this.getBlockHash(height);
+        this.logger.log(
+            `Processing block at height ${height}, hash ${blockHash}`,
+        );
+        const block = await this.getBlock(blockHash, 2);
+        for (const txn of block.tx) {
```
theanmolsharma, line 175: Why not just skip the first transactions? (first transaction is always coinbase transaction)
Then we can remove all the extra logic to handle coinbase transactions

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    private async parseTransaction(
+        txn: BlockTransaction,
+        blockHash: string,
+        blockHeight: number,
+    ): Promise<Transaction> {
+        const inputs: TransactionInput[] = await Promise.all(
+            txn.vin.map(
+                async (input) => await this.parseTransactionInput(input),
+            ),
```
theanmolsharma, line 194: pretty sure you can do something like this
```suggestion
            txn.vin.map(this.parseTransactionInput),
```

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+            scriptPubKey: txnOutput.scriptPubKey.hex,
+            value: txnOutput.value,
+        };
+    }
+
+    async request(body: RPCRequestBody): Promise<any> {
+        const { rpcUser, rpcPass, rpcPort, rpcHost } = this.config;
+        try {
+            const response = await axios.post(
+                `http://${rpcHost}:${rpcPort}/`,
```
theanmolsharma, line 252: add a private read-only property called url or something and set it inside the constructor.

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    private parseTransactionOutput(txnOutput: Output): TransactionOutput {
+        return {
+            scriptPubKey: txnOutput.scriptPubKey.hex,
+            value: txnOutput.value,
+        };
+    }
+
+    async request(body: RPCRequestBody): Promise<any> {
+        const { rpcUser, rpcPass, rpcPort, rpcHost } = this.config;
```
theanmolsharma, line 249: instead of copying this again from `configService` why not use the service here to get it?
so we can do something like
```ts
                {
                     auth: {
                         username: this.configService.get<string>('bitcoinCore.rpcUser'),
                         password: this.configService.get<string>('bitcoinCore.rpcPass'),
                     },
                 },
```

src/block-data-providers/block-provider.module.ts
```diff
@@ -18,11 +20,24 @@ import { IndexerService } from '@/indexer/indexer.service';
...
-                    operationStateService,
-                );
+                switch (configService.get<ProviderType>('providerType')) {
+                    case ProviderType.ESPLORA:
+                        return new EsploraProvider(
+                            configService,
+                            indexerService,
+                            operationStateService,
+                        );
+
```
theanmolsharma, line 30: nit, remove additional new lines

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: I see that you are already logging inside `this.request` so I don't get why we are doing it again here.

package-lock.json
theanmolsharma, line 1: Revert this

src/block-data-providers/bitcoin-core/interfaces.ts
```diff
@@ -0,0 +1,62 @@
...
+    vout: TransactionOutput[];
+    blockHeight: number;
+    blockHash: string;
+};
+
+export interface RPCRequestBody {
+    jsonrpc: string;
+    method: string;
+    id: string;
+    params: unknown[];
```
theanmolsharma, line 61: `unknown[]` is bad. AFAIK params can only be strings, numbers or booleans

src/block-data-providers/bitcoin-core/provider.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                    return 'regtest';
+                }
+                return null;
+            },
+        };
+        provider = new BitcoinCoreProvider(
+            fakeConfigService as unknown as ConfigService,
+            {} as unknown as IndexerService,
+            {} as unknown as OperationStateService,
+        );
```
theanmolsharma, line 33: Please use the standard way of testing using `Test.createTestingModule`. See [this](https://github.com/Bitshala-Incubator/silent-pay-indexer/blob/main/src/indexer/indexer.service.spec.ts) for reference

theanmolsharma (changes requested):

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: I don't think it is necessary, since it is only used in a single function, i have localised such utility to the particular function.

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: Good catch

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: I believe individual preference is at war here. I  rather prefer all config data to be be initialised at the constructor, this helps me to localise any alteration that might be done which pertains to configurations.

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,278 @@
...
+        blockHeight: number,
+    ): Promise<Transaction> {
+        const inputs: TransactionInput[] = await Promise.all(
+            txn.vin.map(
+                async (input) => await this.parseTransactionInput(input),
+            ),
+        );
+        const outputs: TransactionOutput[] = txn.vout.map((output) =>
+            this.parseTransactionOutput(output),
+        );
```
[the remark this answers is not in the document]
aruokhai: Not sure that would would work due to the reference of `this`

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: Sure it would be, but how certain are we that it would be returned in order. The RPC doc states 'returns an Object with information about block <hash> and information about each transaction'. I believe having a safety check does no harm.

aruokhai:
@theanmolsharma , i recently  found out that from version 23 , we have a new verbose type  3  `If verbosity is 3, returns an Object with information about block <hash> and information about each transaction, including prevout information for inputs (only for unpruned blocks in the current best chain).,` which i think provides what we need to avoid excessive calls, i would create an issue to implement that

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: We can pass `this` like this

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: even if the utility is localized, we are creating an additional string that is instantiated multiple times even though its value is not changing. Since `request` can be called a lot, I feel what I suggested is an optimisation.

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: I think we already discussed this and agreed to remove the coinbase check and skip the first transaction

package-lock.json
theanmolsharma, line 1: I think this didn't get reverted properly

theanmolsharma (changes requested):
Before I review futher, please rebase on main and force push

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,237 @@
...
+                },
+            );
+            return response.data.result;
+        } catch (error) {
+            this.logger.error(
+                `Request to BitcoinCore failed!\nRequest:\n${JSON.stringify(
+                    body,
+                )}\nError:\n${error.message}`,
+            );
+            throw error;
```
theanmolsharma, line 234: here we should handle error properly. If it's an `AxiosError` we should log the response.

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    public async processBlock(height: number): Promise<Transaction[]> {
+        const parsedTransactionList: Transaction[] = [];
+        const blockHash = await this.getBlockHash(height);
+        this.logger.debug(
+            `Processing block at height ${height}, hash ${blockHash}`,
+        );
+        const block = await this.getBlock(blockHash, 2);
+        const excludedCoinbaseTxns = block.tx.slice(1);
```
theanmolsharma, line 144: slice is bad, we are unnecessarily creating another copy of the transactions.
Just use a simple for loop like `for(let i = 1; i < block.tx.length; i++)`

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -0,0 +1,237 @@
...
+        for (const txn of excludedCoinbaseTxns) {
+            const parsedTransaction = await this.parseTransaction(
+                txn,
+                block.hash,
+                block.height,
+            );
+            parsedTransactionList.push(parsedTransaction);
+        }
+        return parsedTransactionList;
```
theanmolsharma, line 153: I see that here you are creating an array of transactions and returning to the caller. 
Then in the caller, you are iterating over the array and indexing the transactions.
I feel that you should do the indexing right here.

src/configuration.model.ts
```diff
@@ -64,6 +64,24 @@ class EsploraConfig {
...
+    @IsString()
+    rpcPass: string;
+
+    @IsNotEmpty()
+    @IsString()
+    rpcUser: string;
+
+    @IsNotEmpty()
+    @IsString()
+    rpcPort: string;
```
theanmolsharma, line 82: This should be a number. Please add additional validation checks like min and max

theanmolsharma (changes requested):

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: This is done that way to ensure consistency, that is  when error happens in processing a block, the database state doesn't remain inconsistent by having some transactions in a block processed

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: This would be handled in the axios retry task, that is why i did not include it

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: yes, I get it but in future, we want the ability to be able to resume processing from a partially processed block. This would require storing the index of the last processed tx in the block.

aruokhai: personally i would have liked if the index could be done in a blockwise fashion, so there can never be an inconsistent state. With my current impl, what could make an incosistent state is  this `await this.transactionsService.saveTransaction(transaction);`. If saving all blocks could be done atomically in the DB, there would never be an inconsistent state

theanmolsharma: I hear you. IMHO:
- indexing a full block is a costly operation right now because we are making a lot of network requests
- the chances of failure is high because there are a lot of network requests

that's why I don't feel that indexing a block should be an atomic operation

theanmolsharma: But in your followup PR, if use verbosity 3, then there will only be one network call, and the chances of failure will be low.
So when we make that change, the constraint that block indexing should be atomic will make sense

src/configuration.model.ts
```diff
@@ -64,6 +64,25 @@ class EsploraConfig {
...
+    rpcPass: string;
+
+    @IsNotEmpty()
+    @IsString()
+    rpcUser: string;
+
+    @IsInt()
+    @Min(1)
+    @Max(65535)
+    rpcPort: string;
```
theanmolsharma, line 83:
```suggestion
    rpcPort: number;
```

src/block-data-providers/bitcoin-core/provider.ts
theanmolsharma: @aruokhai I think you missed this

theanmolsharma (changes requested):
Final set of changes.
Please cleanup the commit history by squashing unnecessary commits.

I think there should only be 2 commits:
1. add config
2. add the provider

You also need to rebase on upstream main again

theanmolsharma (approved):
Really nice work @aruokhai
Thanks for putting in the effort to do this!!

theanmolsharma (approved):
ACK 233037ef27c3a7c88f8378476f960b7c68744439

Issue #17, opened 2024-06-25
Implement Retry Mechanism for Failed Requests in EsploraProvider


theanmolsharma:
In the current implementation of the `EsploraProvider` class, the `request` method does not retry the request if a server error occurs.

We should implement a retry mechanism in the `request` method. The retry mechanism should handle server errors (HTTP status codes 5xx) and network errors, and should have a limit on the number of retry attempts to prevent infinite loops.

aruokhai:
Taking this

theanmolsharma:
Merged in #20

PR #20, opened 2024-07-24
Providers Axios Retry Mechanism

aruokhai:
### Objective
It is paramount for a service that makes multiple Http calls such as Esplora Provider to have an Error Retry Mechanism, which is what is being rectified by this pull request

### Changes
- Implemented a custom  error retry logic for http calls  with exponential delay logic 
-  Modified  how  error logs are thrown During Http calls

### Scope Of Change 
Crittical , as it affects the fetching of data.

Solves #17

### DIFF
--- a/src/block-data-providers/esplora.provider.ts
+++ b/src/block-data-providers/esplora.provider.ts
@@ -1,13 +1,13 @@
 import { Injectable, Logger, OnApplicationBootstrap } from '@nestjs/common';
 import { OperationStateService } from '@/operation-state/operation-state.service';
 import { BaseBlockDataProvider } from '@/block-data-providers/base-block-data-provider.abstract';
-import axios, { AxiosError, AxiosRequestConfig } from 'axios';
 import { TAPROOT_ACTIVATION_HEIGHT } from '@/common/constants';
 import { ConfigService } from '@nestjs/config';
 import { BitcoinNetwork } from '@/common/enum';
 import { URL } from 'url';
 import { Cron, CronExpression } from '@nestjs/schedule';
 import { IndexerService, TransactionInput } from '@/indexer/indexer.service';
+import { AxiosRetryConfig, makeRequest } from '@/common/request';
 
 type EsploraOperationState = {
     currentBlockHeight: number;
@@ -64,6 +64,7 @@ export class EsploraProvider
     protected readonly logger = new Logger(EsploraProvider.name);
     protected readonly operationStateKey = 'esplora-operation-state';
     private readonly baseUrl: string;
+    private retryConfig: AxiosRetryConfig;
     private isSyncing = false;
     private readonly batchSize: number;
 
@@ -91,6 +92,9 @@ export class EsploraProvider
         this.baseUrl = new URL(
             `${this.configService.get<string>('esplora.url')}${pathPrefix}`,
         ).toString();
+
+        this.retryConfig =
+            this.configService.get<AxiosRetryConfig>('app.requestRetry');
     }
 
     async onApplicationBootstrap() {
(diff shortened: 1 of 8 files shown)


package.json
theanmolsharma, line 1: As discussed, please remove this dependency and implement custom retry logic

src/configuration.model.ts
```diff
@@ -57,6 +57,9 @@ class EsploraConfig {
         require_host: true,
     })
     url: string;
+    @IsInt()
```
theanmolsharma, line 60: nit
```suggestion
    url: string;

    @IsInt()
```

theanmolsharma (changes requested):

src/common/request.ts
```diff
@@ -0,0 +1,106 @@
...
+        return false;
+    }
+
+    // Prevents retrying timed out & cancelled requests
+    if (CODE_EXCLUDE_LIST.includes(error.code)) {
+        return false;
+    }
+
+    return true;
+};
```
theanmolsharma, line 101:
```suggestion
export const isNetworkError = (error): boolean => {
    return !(
        error.response ||
        !error.code ||
        ['ERR_CANCELED', 'ECONNABORTED'].includes(error.code)
    );
};
```

theanmolsharma (changes requested):

src/common/request.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+};
+
+const exponentialDelay = (
+    retryNumber: number,
+    retryConfig: AxiosRetryConfig,
+): Promise<void> => {
+    const calculatedDelay = 2 ** retryNumber * retryConfig.delay;
+    const delay = calculatedDelay;
+    const randomSum = delay * 0.2 * Math.random(); // 0-20% of the delay
```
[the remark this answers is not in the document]
aruokhai: It specifically is to add a little bit of jitter to make the calls less systematic which helps to avoids synchronised failures and distribute retries over time.  For example  If multiple request  attempt to retry a failed operation at the same exact time which is possible for esplora , they may all encounter the same underlying issue, leading to a cascading failure. Jitter introduces randomness into the retry timing, helping to spread out retries and reduce the likelihood of synchronized failures.

src/common/request.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const axiosErrorResponse = (error: AxiosError) =>
+    error.response?.data || error.message;
+
+const exponentialDelay = (
+    retryNumber: number,
+    retryConfig: AxiosRetryConfig,
+): Promise<void> => {
+    const calculatedDelay = 2 ** retryNumber * retryConfig.delay;
+    const delay = calculatedDelay;
```
theanmolsharma, line 15:
```suggestion
    const delay = 2 ** retryNumber * retryConfig.delay;
```

src/common/request.ts
```diff
@@ -0,0 +1,81 @@
...
+
+                await exponentialDelay(count, retryConfig);
+
+                continue;
+            }
+
+            logger.error(`unknown error encountered ${error}`);
+            throw error;
+        }
+    }
```
theanmolsharma, line 67: Axios can sometimes throw [AggregateError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError), we should also handle them here.
doing something like
```ts
        if (error instanceof AggregateError) {
            logger.error(
                `Aggregate Error encountered: ${error.message}`,
                error.stack,
            );
            for (const cause of error.errors) {
                logger.error(`Cause: ${cause.message}`, cause.stack);
            }
        }
```

src/block-data-providers/esplora.provider.ts
```diff
@@ -258,4 +224,8 @@ export class EsploraProvider
             url: `${this.baseUrl}/tx/${txid}`,
         });
     }
+
+    private async request(request_config: AxiosRequestConfig): Promise<any> {
+        return makeRequest(request_config, this.retryConfig, this.logger);
+    }
```
theanmolsharma, line 230: Instead of this, we should directly call `makeRequest` wherever `request` is called

theanmolsharma (changes requested):

src/common/request.ts
aruokhai: Funny how i didn't catch this :smile:

theanmolsharma: nw dude, that's what reviews are for

PR #22, opened 2024-08-16
Added `batchSize` configuration option for parallelizing requests in EsploraProvider

notTanveer:
Added support for batch processing in EsploraProvider by introducing a batchSize configuration option. Requests in the processBlock method are now parallelized using Promise.all() when batchSize > 1, improving efficiency. An upper limit for batchSize(100) is enforced to prevent overloading. This change optimizes block processing while respecting API rate limits.

### DIFF
--- a/config/dev.config.yaml
+++ b/config/dev.config.yaml
@@ -11,3 +11,4 @@ app:
 providerType: ESPLORA
 esplora:
   url: https://blockstream.info
+  batchSize: 5
--- a/src/block-data-providers/esplora.provider.ts
+++ b/src/block-data-providers/esplora.provider.ts
@@ -65,6 +65,7 @@ export class EsploraProvider
     protected readonly operationStateKey = 'esplora-operation-state';
     private readonly baseUrl: string;
     private isSyncing = false;
+    private readonly batchSize: number;
 
     constructor(
         private readonly configService: ConfigService,
@@ -72,6 +73,8 @@ export class EsploraProvider
         operationStateService: OperationStateService,
     ) {
         super(indexerService, operationStateService);
+
+        this.batchSize = this.configService.get<number>('esplora.batchSize');
 
         let pathPrefix;
         switch (this.configService.get<BitcoinNetwork>('app.network')) {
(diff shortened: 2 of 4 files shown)


src/block-data-providers/esplora.provider.ts
```diff
@@ -160,12 +163,36 @@ export class EsploraProvider
 
     private async processBlock(height: number, hash: string) {
         const txids = await this.getTxidsForBlock(hash);
+        const batchSize = this.configService.get<number>('app.batchSize') || 1;
+
+        const MAX_BATCH_SIZE = 100;
```
theanmolsharma, line 168: This should be enforced through config validation

src/block-data-providers/esplora.provider.ts
```diff
@@ -160,12 +163,36 @@ export class EsploraProvider
 
     private async processBlock(height: number, hash: string) {
         const txids = await this.getTxidsForBlock(hash);
+        const batchSize = this.configService.get<number>('app.batchSize') || 1;
```
theanmolsharma, line 166: if you have created a class property `batchSize`, why are you not using it?

src/block-data-providers/esplora.provider.ts
```diff
@@ -160,12 +163,36 @@ export class EsploraProvider
...
+
+        const MAX_BATCH_SIZE = 100;
+        const effectiveBatchSize = Math.min(batchSize, MAX_BATCH_SIZE);
+
+        if (effectiveBatchSize > 1) {
```
theanmolsharma, line 171: redundant check

src/block-data-providers/esplora.provider.ts
```diff
@@ -160,12 +163,36 @@ export class EsploraProvider
...
+            for (let i = 0; i < txids.length; i += effectiveBatchSize) {
```
theanmolsharma, line 172: you should skip the first transaction because it's coinbase transaction

theanmolsharma (changes requested):
You also need to update the default and test config files and the `src/configuration.model.ts` file

config/dev.config.yaml
```diff
@@ -8,6 +8,7 @@ db:
 app:
   port: 3000
   network: main
+  batchSize: 100
```
theanmolsharma, line 11: set it to something low. Maybe 5 or 3

theanmolsharma: also because this option is specific to esplora, it should be a part of esplora config not under the app

src/block-data-providers/esplora.provider.ts
```diff
@@ -160,28 +163,30 @@ export class EsploraProvider
 
     private async processBlock(height: number, hash: string) {
         const txids = await this.getTxidsForBlock(hash);
+        const batchSize = this.batchsize;
```
theanmolsharma, line 166: why do this? why not directly use `this.batchSize`?

theanmolsharma (changes requested):
Please do the following:
- setup prettier and make sure the changes you submit are properly formatted and passing the lint and format check
- squash all commits into a single commit. We preserve entire commit history upon merge so having unnecessary commits is bad.
- sign all commits. We can only merge signed and verified commit. If you haven't setup commit signing, you can go through [this](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

src/block-data-providers/esplora.provider.ts
```diff
@@ -161,26 +164,27 @@ export class EsploraProvider
...
-                witness: input.witness,
-            }));
-            const vout = tx.vout.map((output) => ({
-                scriptPubKey: output.scriptpubkey,
-                value: output.value,
-            }));
-
-            await this.indexTransaction(txid, vin, vout, height, hash);
+        for (let i = 1; i < txids.length; i += this.batchSize) {
+            const batch = txids.slice(i, i + this.batchSize);
```
theanmolsharma, line 168: slice is a costly operation, I bet there is a way to remove it.
Think of some more performance optimisations that you can do, like removing `map` etc

notTanveer: should i use a different for loop for slice operation?

theanmolsharma: nvm this is fine

theanmolsharma (changes requested):
CI is failing because ESlint and prettier checks are failing. You need to fix it before we can merge.
You also didn't rebase on upstream main. You can run the following command to do it.
```bash
git pull --rebase upstream main
```

theanmolsharma (approved):

PR #32, opened 2024-09-02
Improve Bitcoin Core Provider Block Fetching Functionality

aruokhai:
### Objective

A new Verbosity Level of 3 was added to `Bitcoin Core` RPC `getblock` Method from V23, this allows the retrieval of the Pvev Output details  of Transactions Inputs, which helps to reduce the number of calls made to Bitcoin Core.  

### Changes
- Added a new method call `getNetworkInfo`, which offers the dynamic retrieval of bitcoincore version .
- Added Logic to ensure that Verbosity Level 3 is utilised when Bitcoin core version number >= 23.

### Scope Of Change
The scope is quite concerned with Bitcoin Core Provider and it is a critical change because it affects the speed of block retrieval from bitcoin core.


src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -243,4 +273,8 @@ export class BitcoinCoreProvider
     private convertToSatoshi(amount: number): number {
         return currency(amount, { precision: 8 }).multiply(SATS_PER_BTC).value;
     }
+
+    private hasFullVerboseInput(version: number): boolean {
+        return version >= BITCOIN_CORE_FULL_VERBOSITY_VERSION;
+    }
```
theanmolsharma, line 279: Let's change it to this.

```suggestion
    private versionToVerbosity(version: number): 2 | 3 {
        return version >= BITCOIN_CORE_FULL_VERBOSITY_VERSION ? 3 : 2;
    }
```

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -185,13 +208,20 @@ export class BitcoinCoreProvider
     private async parseTransactionInput(
         txnInput: Input,
     ): Promise<TransactionInput> {
-        const prevTransaction = await this.getRawTransaction(
-            txnInput.txid,
-            true,
-        );
+        let prevOutScript = '';
```
theanmolsharma, line 211:
```suggestion
        let prevOutScript: string;
```

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -86,9 +91,17 @@ export class BitcoinCoreProvider
             return;
         }
 
+        const networkInfo = await this.getNetworkInfo();
+        const verbosityLevel = this.hasFullVerboseInput(networkInfo.version)
+            ? 3
+            : 2;
```
theanmolsharma, line 97: Let's create a class variable called `verbosityLevel` and set it inside `onApplicationBootstrap` instead of `sync`
The version of node software won't change in-between calls so fetching it over and over is useless IMHO.

theanmolsharma (changes requested):

src/block-data-providers/bitcoin-core/provider.ts
aruokhai: I understand, that is why i put it in the sync function, because there surely cant be a change of bitcoind version while in sync but it is logically possible to be switched after sync. Logically possible because all the config depends on is the RPC details.

theanmolsharma: What do you mean by in-sync and after sync?

theanmolsharma: I changed my mind, just leave it in there for now

theanmolsharma (approved):

Issue #19, opened 2024-06-25
Create and endpoint to serve silent block for specific hash or height


theanmolsharma:
There is no dedicated endpoint to serve a silent block for a specific block hash or height.

This issue is to track the development of two new endpoints that serve a silent block for a given block hash and height. The endpoints should accept the block hash and the block height respectively. They should return the corresponding silent block in the format specified in the [`design.md`](https://github.com/Bitshala-Incubator/silent-pay-indexer/blob/main/docs/design.md#specification-for-silent-block) document.

This enhancement will allow clients to fetch all tweaks for a specific block.

seekersoftec:
Taking this

theanmolsharma:
Merged in #41

PR #35, opened 2024-09-06
Implement Silent Blocks Endpoint

seekersoftec:
**Objective:**
Create a new endpoint to serve silent blocks based on specific hash or height, and enhance the project's documentation using Swagger.

**Changes:**
- **Silent Blocks Endpoint:** Implemented a RESTful endpoint to retrieve silent blocks by providing either a specific hash or height.
- **Swagger Documentation:** Integrated Swagger to generate interactive API documentation, making it easier for developers to understand and use the endpoint.
- **SQL Loader for Testing:** Added an SQL loader for testing purposes, facilitating end-to-end testing of the silent blocks endpoint.

**Scope of Change:**
This PR primarily focuses on adding the new endpoint, implementing Swagger documentation, and improving the e2e testing suite.

Solves #19

### DIFF
--- a/src/block-data-providers/base-block-data-provider.abstract.ts
+++ b/src/block-data-providers/base-block-data-provider.abstract.ts
@@ -6,7 +6,7 @@ import {
     TransactionOutput,
 } from '@/indexer/indexer.service';
 
-export abstract class BaseBlockDataProvider {
+export abstract class BaseBlockDataProvider<OperationState> {
     protected abstract readonly logger: Logger;
     protected abstract readonly operationStateKey: string;
 
@@ -31,7 +31,7 @@ export abstract class BaseBlockDataProvider {
         );
     }
 
-    async getState(): Promise<Record<string, any>> {
+    async getState(): Promise<OperationState> {
         return (
             await this.operationStateService.getOperationState(
                 this.operationStateKey,
@@ -39,7 +39,7 @@ export abstract class BaseBlockDataProvider {
         )?.state;
     }
 
-    async setState(state: Record<string, any>): Promise<void> {
+    async setState(state: OperationState): Promise<void> {
         await this.operationStateService.setOperationState(
             this.operationStateKey,
             state,
--- a/src/block-data-providers/bitcoin-core/provider.ts
+++ b/src/block-data-providers/bitcoin-core/provider.ts
@@ -31,7 +31,7 @@ import { AxiosRetryConfig, makeRequest } from '@/common/request';
 
 @Injectable()
 export class BitcoinCoreProvider
-    extends BaseBlockDataProvider
+    extends BaseBlockDataProvider<BitcoinCoreOperationState>
     implements OnApplicationBootstrap
 {
     protected readonly logger = new Logger(BitcoinCoreProvider.name);
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -15,7 +15,7 @@ import { Cron, CronExpression } from '@nestjs/schedule';
 
 @Injectable()
 export class EsploraProvider
-    extends BaseBlockDataProvider
+    extends BaseBlockDataProvider<EsploraOperationState>
     implements OnApplicationBootstrap
 {
     protected readonly logger = new Logger(EsploraProvider.name);


package.json
```diff
@@ -33,7 +33,7 @@
     "@nestjs/platform-express": "^10.3.7",
     "@nestjs/platform-socket.io": "^10.3.7",
     "@nestjs/schedule": "^3.0.0",
-    "@nestjs/swagger": "^7.3.1",
+    "@nestjs/swagger": "^7.4.0",
```
theanmolsharma, line 37: any reason for updating the version here?

package.json
```diff
@@ -43,6 +43,7 @@
     "rxjs": "^7.2.0",
     "secp256k1": "^5.0.0",
     "sqlite3": "^5.1.7",
+    "swagger-ui-express": "^5.0.1",
```
theanmolsharma, line 47: Why are adding this? I don't think it's used

src/main.ts
```diff
@@ -1,15 +1,27 @@
 import { NestFactory } from '@nestjs/core';
 import { AppModule } from '@/app.module';
 import { ConfigService } from '@nestjs/config';
+import { NestExpressApplication } from '@nestjs/platform-express';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
+import * as packageJson from 'package.json';
```
theanmolsharma, line 6: I don't like the idea of importing from a json file. This would require allowing json imports all across the app which is a bad thing IMHO. Please remove this and just duplicate the description and version for now.

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Encodes a variable integer (varInt) into a Buffer.
+     * @param value - The integer value to encode.
+     * @returns A Buffer containing the encoded varInt.
+     */
+    private encodeVarInt(value: number): Buffer {
+        const buffer = Buffer.alloc(1);
+        buffer.writeUInt8(value);
+        return buffer;
+    }
```
theanmolsharma, line 145: this is not the right way to encode varint. Please follow standard bitcoin varint encoding and decoding

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Decodes a variable integer (varInt) from a Buffer.
+     * @param buffer - The buffer containing the encoded varInt.
+     * @param offset - The offset in the buffer where the varInt starts.
+     * @returns The decoded integer value.
+     */
+    private decodeVarInt(buffer: Buffer, offset: number): number {
+        return buffer.readUInt8(offset);
+    }
```
theanmolsharma, line 155: this is not right as mentioned in previous comment

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Encodes an array of transactions into a silent block buffer.
+     * @param transactions - The transactions to encode.
+     * @returns A Buffer representing the encoded silent block.
+     */
+    private encodeSilentBlock(transactions: Transaction[]): Buffer {
+        const silentBlockType = SILENT_PAYMENT_BLOCK_TYPE;
+        const transactionBuffers: Buffer[] = [];
+
+        transactions.forEach((tx) => {
```
theanmolsharma, line 20: instead of `forEach` you can directly use `map`

theanmolsharma (changes requested):
Regarding 298e8b3d474e2c7a1155fa640467b535d09ec0d6
- we don't have any e2e tests now so there is no point in this commit
- even if we wrote the test, we won't load transactions like this. We will create the transactions using a bitcoin node and then the indexer should pick it up.

Please remove this commit for now and shelf this change

package.json
seekersoftec: I guess this happened when I tried to re-install Swagger without knowing it was here already

src/main.ts
seekersoftec: Okay

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,219 @@
...
+        if (firstByte < 0xfd) {
+            return { value: firstByte, bytesRead: 1 };
+        } else if (firstByte === 0xfd) {
+            return { value: buffer.readUInt16LE(offset + 1), bytesRead: 3 };
+        } else if (firstByte === 0xfe) {
+            return { value: buffer.readUInt32LE(offset + 1), bytesRead: 5 };
+        } else {
+            return { value: buffer.readBigUInt64LE(offset + 1), bytesRead: 9 };
+        }
+    }
```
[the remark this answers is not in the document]
seekersoftec: Done, thanks.

src/silent-blocks/silent-blocks.service.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+                    {
+                        pubKey: '5af9436eb515871413a913f24290ba003eab7a64b9c62cb72051599a383de8ff',
+                        value: 26940,
+                        vout: 0,
+                    },
+                    {
+                        pubKey: 'e7a3ab33572f47b074878c749a5f9a5d7e026f10b6ffcab1388e9356c07828fa',
+                        value: 4400,
+                        vout: 1,
+                    },
```
aruokhai, line 69: move to a different file

src/silent-blocks/silent-blocks.service.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+                isSpent: false,
+            },
+        ];
+
+        // Mock the TransactionsService to return the mock transactions
+        jest.spyOn(
+            transactionsService,
+            'getTransactionByBlockHeight',
+        ).mockResolvedValue(mockTransactions);
+
```
aruokhai, line 80: `mockImplementation` can be used and the need mock Values can be gotten through the combination of map and simple logic.

src/silent-blocks/silent-blocks.service.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+                outputs: [
+                    {
+                        pubKey: 'ca80a975d09b84b385e39d779293121b9a7c80942d824d2496f098e3f077ced3',
+                        value: 24771038,
+                        vout: 1,
+                    },
+                ],
+                isSpent: false,
+            },
+        ];
```
aruokhai, line 137: same here

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,163 @@
...
+                        valueBuffer,
+                        pubKeyBuffer,
+                        voutBuffer,
+                    ]);
+                }),
+            );
+
+            const scanTweakBuffer = Buffer.from(tx.scanTweak, 'hex');
+
+            return Buffer.concat([
```
aruokhai, line 44: can you move this to a new file in the common called `silent-block`

aruokhai:
@seekersoftec , please close this pull request

PR #36, opened 2024-09-06
E2E test

notTanveer:
test for rpc block height and indexed block height.


aruokhai:
Hey @notTanveer, very good effort. I reckon we should be very particularly concerned with external APIs, that is , e2e test should act like a client making requests to the indexer and should **assert** a given response. The overaching goal is to treat the indexer as a black box.

PR #38, opened 2024-09-11
Modified Unit Test Of Indexer Service

aruokhai:
Indexer Service Test Modified to utilize In Memory Sqlite DB


src/indexer/indexer.fixture.ts
theanmolsharma, line 1: Although this is just mock test data, I think we should preserve the constraints like the length of `txid`. Please update all txids to be 64 characters long (32 bytes).

src/indexer/indexer.service.spec.ts
```diff
@@ -30,24 +41,27 @@ describe('IndexerService', () => {
...
             0,
             '0000000000000000000000000000000000000000000000000000000000000000',
         );
 
-        if (testData.scanTweak) {
-            expect(saveMock).toHaveBeenCalled();
-            expect(saveMock.mock.calls[0][0].scanTweak).toBe(
-                testData.scanTweak,
-            );
+        const transaction_index = await repository.findOne({
```
theanmolsharma, line 53: nit
```suggestion
        const transactionEntity = await repository.findOne({
```

theanmolsharma (changes requested):

theanmolsharma (approved):

PR #40, opened 2024-09-13
Improve processBlock Method for Resumable Indexing in EsploraProvider

notTanveer:
This PR enhances the processBlock method in the EsploraProvider class by making the block indexing process resumable. It introduces tracking for the last successfully indexed transaction within a block, allowing the indexer to resume from where it left off if interrupted, instead of reindexing the entire block. This improvement increases efficiency and ensures that no redundant work is performed during the indexing process.

### DIFF
--- a/src/block-data-providers/esplora.provider.ts
+++ b/src/block-data-providers/esplora.provider.ts
@@ -12,6 +12,7 @@ import { AxiosRetryConfig, makeRequest } from '@/common/request';
 type EsploraOperationState = {
     currentBlockHeight: number;
     indexedBlockHeight: number;
+    lastProcessedTxIndex: number;
 };
 
 type EsploraTransactionInput = {
@@ -114,6 +115,7 @@ export class EsploraProvider
                     BitcoinNetwork.MAINNET
                         ? TAPROOT_ACTIVATION_HEIGHT - 1
                         : 0,
+                lastProcessedTxIndex: 0, // we dont take coinbase txn in account
             };
             await this.setState(state);
         }
@@ -140,7 +142,7 @@ export class EsploraProvider
             }
 
             for (
-                let height = state.indexedBlockHeight + 1;
+                let height = state.indexedBlockHeight;
                 height <= tipHeight;
                 height++
             ) {
@@ -150,9 +152,6 @@ export class EsploraProvider
                 );
 
                 await this.processBlock(height, blockHash);
-
-                state.indexedBlockHeight = height;
-                await this.setState(state);
             }
         } finally {
             this.isSyncing = false;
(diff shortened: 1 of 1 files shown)


src/block-data-providers/esplora.provider.ts
```diff
@@ -162,33 +162,59 @@ export class EsploraProvider
     }
 
     private async processBlock(height: number, hash: string) {
+        const state = await this.getState();
+        const lastProcessedTxIndex = state?.lastProcessedTxIndex ?? 0;
```
theanmolsharma, line 166: Assume that it's defined
```suggestion
        const lastProcessedTxIndex = state.lastProcessedTxIndex;
```

src/block-data-providers/esplora.provider.ts
```diff
@@ -162,33 +162,59 @@ export class EsploraProvider
...
+
         const txids = await this.getTxidsForBlock(hash);
 
-        for (let i = 1; i < txids.length; i += this.batchSize) {
+        for (
+            let i = lastProcessedTxIndex;
```
theanmolsharma, line 171: I think this should be `lastProcessedTxIndex + 1`. If the last processed index was `10`, we want to start from `11`. The default vault should be `-1`

src/block-data-providers/esplora.provider.ts
```diff
@@ -162,33 +162,59 @@ export class EsploraProvider
...
+            } catch (error) {
+                this.logger.error(
+                    `Error processing transactions in block at height ${height}, hash ${hash}: ${error.message}`,
+                );
+                throw error;
+            }
         }
+
+        state.lastProcessedTxIndex = 0;
+        await this.setState(state);
```
theanmolsharma, line 217: I don't get why we are setting it to 0 here. We should do anything here.

src/block-data-providers/esplora.provider.ts
theanmolsharma, line 1: You also need to edit `EsploraOperationState` and add `lastProcessedTxIndex: number`
And initialise the state in `onApplicationBootstrap` when the state is not defined

theanmolsharma (approved):

src/block-data-providers/esplora.provider.ts
notTanveer: once we are done processing the entire block. Reset the last processed transaction index .

theanmolsharma: why do we need to reset after finishing the processing?

src/block-data-providers/esplora.provider.ts
```diff
@@ -162,32 +164,55 @@ export class EsploraProvider
     }
 
     private async processBlock(height: number, hash: string) {
+        const state = await this.getState();
+        const lastProcessedTxIndex = state.lastProcessedTxIndex;
```
theanmolsharma, line 168: I think you can just use `state.lastProcessedTxIndex` below

src/block-data-providers/esplora.provider.ts
```diff
@@ -110,6 +111,7 @@ export class EsploraProvider
                     BitcoinNetwork.MAINNET
                         ? TAPROOT_ACTIVATION_HEIGHT - 1
                         : 0,
+                lastProcessedTxIndex: -1,
```
theanmolsharma, line 114: now that i think about it, the default value should be `0`, not `-1`
This is because we want to skip the coinbase transaction

theanmolsharma (changes requested):

PR #41, opened 2024-09-15
feat: added endpoint to serve silent block for specific hash or height

theanmolsharma:


src/common/common.ts
```diff
@@ -117,3 +117,33 @@ export const extractPubKeyFromScript = (
...
+        buffer.writeUInt32LE(value, offset + 4);
+        return offset + 5;
+    } else {
+        buffer.writeUInt8(0xff, offset);
+        buffer.writeBigUInt64LE(BigInt(value), offset + 8);
+        return offset + 9;
+    }
+};
+
+export const varIntSize = (value: number): number => {
```
aruokhai, line 144: I believe `varIntSize` can also be used in `encodeVarInt` as the final return statement `return offset + varIntSize(value)`

src/silent-blocks/silent-blocks.service.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    it.each(silentBlockEncodingFixture)(
+        'should encode a silent block correctly by block hash: $blockHash',
+        async ({ transactions, blockHash, encodedBlockHex }) => {
+            await transactionRepository.save(transactions);
+
+            const encodedBlock = await service.getSilentBlockByHash(blockHash);
+
+            expect(encodedBlock.toString('hex')).toEqual(encodedBlockHex);
+        },
+    );
```
aruokhai, line 66: It would be splendid to also test the `Decoded Value`. and maybe the `DecodeSilentBlock fn` can be put in common for E2E

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -0,0 +1,58 @@
...
+                cursor = block.writeBigUInt64BE(BigInt(output.value), cursor);
+                cursor += block.write(output.pubKey, cursor, 32, 'hex');
+                cursor = block.writeUInt32BE(output.vout, cursor);
+            }
+
+            cursor += block.write(tx.scanTweak, cursor, 33, 'hex');
+        }
+
+        return block;
+    }
```
aruokhai, line 43: This can be moved to a separate file and maybe `decodeSilentBlock  fn()` can also be included, which would allow a comprehensive test for the encoding.

src/silent-blocks/silent-blocks.service.spec.ts
theanmolsharma: the decode function will be implemented in the sore silent pay library. Once that is done, we can use it inside the tests.
For now, I think it's fine because I did a manual byte-by-byte encoding and used that in the fixture.

src/silent-blocks/silent-blocks.service.ts
theanmolsharma: I don't see merit in moving it to a separate file. We can comprehensively test without moving to a new file.

src/silent-blocks/silent-blocks.service.spec.ts
aruokhai: resolved

src/silent-blocks/silent-blocks.service.ts
aruokhai: resolved

src/common/common.ts
aruokhai: resolved

src/common/common.ts
```diff
@@ -117,3 +117,33 @@ export const extractPubKeyFromScript = (
...
+    value: number,
+    buffer: Buffer,
+    offset = 0,
+): number => {
+    if (value < 0xfd) {
+        buffer.writeUInt8(value, offset);
+        return offset + 1;
+    } else if (value <= 0xffff) {
+        buffer.writeUInt8(0xfd, offset);
+        buffer.writeUInt16LE(value, offset + 2);
```
notTanveer, line 131: shouldn't this be `offset + 1`. we only need to skip the prefix right?

src/silent-blocks/silent-blocks.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+@Controller('silent-block')
+export class SilentBlocksController {
+    constructor(private readonly silentBlocksService: SilentBlocksService) {}
+
+    @Get('height/:blockHeight')
+    async getSilentBlockByHeight(
+        @Param('blockHeight') blockHeight: number,
+        @Res() res: Response,
+    ) {
+        const buffer = await this.silentBlocksService.getSilentBlockByHeight(
```
notTanveer, line 14: how about adding this within a try catch. and log responses for 4xx error. like block not found.

theanmolsharma: in that case we'll just return an empty block

Issue #45, opened 2024-09-20
Implement WebSocket to Stream Silent Blocks


theanmolsharma:
**Description:**
Implement a WebSocket connection to stream silent blocks to the client whenever a new block is found and indexed. This will eliminate the need for the client to keep polling for new blocks.

**Tasks:**
1. Set up a WebSocket server.
2. Implement a method to broadcast new silent blocks to connected clients.
3. Add a function to `src/block-data-providers/base-block-data-provider.abstract.ts` that emits an event whenever a new block is indexed.
4. On receiving the event, the `SilentBlocksService` should create and stream the silent block to all connected clients.
5. Ensure the WebSocket connection is properly managed and closed when clients disconnect.

aruokhai:
@theanmolsharma , can this can be closed ?

theanmolsharma:
Merged in #46

PR #46, opened 2024-09-24
Implement WebSocket to Stream Silent Blocks

seekersoftec:
**Objective(s):**
Implement a WebSocket connection to stream silent blocks to the client whenever a new block is found and indexed. This will eliminate the need for the client to keep polling for new blocks.

**Changes:**
* **WebSocket Server Setup:**
    * Integrated a WebSocket server using the `ws` library.
* **Event Listener:**
    * Add a function to `base-block-data-provider.abstract.ts` that emits an event whenever a new block is indexed.
    * Subscribe to this event in the `SilentBlocksService` and trigger the silent block broadcasting logic when the event is fired.

**Scope of Change:**
This change is **semi-critical**. While it doesn't directly affect core functionality, it significantly enhances the user experience and system efficiency by providing real-time updates. 

solves #45


config/dev.config.yaml
theanmolsharma, line 1: I think this was a mistake. please revert

package.json
```diff
@@ -47,6 +48,7 @@
   },
   "devDependencies": {
     "@nestjs/cli": "^10.4.5",
+    "@nestjs/platform-ws": "^10.4.4",
```
theanmolsharma, line 51: I don't think this should be a dev dependency

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -122,6 +124,7 @@ export class BitcoinCoreProvider
                 }
 
                 state.indexedBlockHeight = height;
+                this.eventEmitter.emit('blockIndexed', height);
```
theanmolsharma, line 127: `'blockIndexed'` should be a const defined in `events.ts` in common directory.
We should call it `INDEXED_BLOCK_EVENT`

theanmolsharma: This should be emitted after the state is set.

src/block-data-providers/block-provider.module.ts
```diff
@@ -7,31 +7,46 @@ import { IndexerService } from '@/indexer/indexer.service';
...
     providers: [
         {
             provide: 'BlockDataProvider',
-            inject: [ConfigService, IndexerService, OperationStateService],
+            inject: [
+                ConfigService,
+                IndexerService,
+                OperationStateService,
+                EventEmitter2,
+                ,
```
theanmolsharma, line 28: nit: 
```suggestion
```

src/silent-blocks/silent-blocks.gateway.ts
```diff
@@ -0,0 +1,40 @@
...
+    implements OnGatewayConnection, OnGatewayDisconnect
+{
+    private readonly logger = new Logger(SilentBlocksGateway.name);
+
+
+    @WebSocketServer() server: Server;
+
+    handleConnection(client: WebSocket) {
+        const remoteAddress = (client as any)._socket.remoteAddress;
+        this.logger.log(`Client connected: ${remoteAddress}`);
```
theanmolsharma, line 22: this should be `debug` level log

src/silent-blocks/silent-blocks.gateway.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    handleDisconnect(client: WebSocket) {
+        const remoteAddress = (client as any)._socket.remoteAddress;
+        this.logger.log(`Client disconnected: ${remoteAddress}`);
```
theanmolsharma, line 27: same here, change to `debug` level log

theanmolsharma (changes requested):
A few more thoughts:
1. I see that you are injecting `SilentBlocksGateway` into `SilentBlocksService`. That is not the standard pattern. Gateways are like Controllers. You don't inject controllers into a service, you inject service into a controller. Therefore, please remove the event handling logic from `SilentBlocksService` and move it to inside of `SilentBlocksGateway`.
2. I also think we should use `socket.io` and leverage the concept of [rooms](https://socket.io/docs/v3/rooms/). Once a client connects to use, we can subscribe them to a `SILENT_BLOCK` room. Then we just need to publish silent block to that room and let socket.io handle it from there. This will eliminate the need of `forEach` loop in `broadcastSilentBlock`

seekersoftec:
Okay, changes noted

config/dev.config.yaml
seekersoftec: Okay

src/common/events.ts
```diff
@@ -0,0 +1 @@
+export const INDEXED_BLOCK_EVENT = 'blockIndexed';
```
theanmolsharma, line 1:
```suggestion
export const INDEXED_BLOCK_EVENT = 'INDEXED_BLOCK_EVENT';
```

theanmolsharma (commented):
I don't think you made all the changes I requested. Please do so before I review it again.

src/silent-blocks/silent-blocks.gateway.ts
```diff
@@ -0,0 +1,37 @@
...
+        this.logger.debug(`Client disconnected: ${remoteAddress}`);
+    }
+
+    // Method to broadcast silent block to all connected clients
+    broadcastSilentBlock(silentBlock: Buffer) {
+        this.server.clients.forEach((client: WebSocket) => {
+            if (client.readyState === WebSocket.OPEN) {
+                client.send(silentBlock.toString('hex')); // Send silent block as hex string
+            }
+        });
```
theanmolsharma, line 35: 1. because everything inside `forEach` is a synchronous operation, we should use a simple `for..of` loop. `forEach` is very slow.
2. I see that you are sending the block as a hex string. This breaks the pattern we use for REST endpoints. We send the block as `Buffer` in REST endpoints. I think we should stick to using `Buffer` here as well for consistency.

theanmolsharma (changes requested):
Please do the following:
1. fix linting issues
2. fix broken unit tests
3. squash all commits into just one commit

aruokhai:
Please Rebase From main

seekersoftec:
> Please Rebase From main

Okay

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -1,12 +1,28 @@
...
+
+    constructor(
+        private readonly transactionsService: TransactionsService,
+        private readonly silentBlocksGateway: SilentBlocksGateway,
+    ) {}
+
+    @OnEvent(INDEXED_BLOCK_EVENT)
+    async handleBlockIndexedEvent(blockHeight: number) {
+        this.logger.debug(`New block indexed: ${blockHeight}`);
+        await delay(1000);
```
aruokhai, line 22: @seekersoftec , kindly research if we could use a  superior way to handle the latency, we could use a while loop with a cut-off time, the cut-off which i believe may be quite unnecessary due to the operational model that requires such SiientBlock to be eventually available .

aruokhai (changes requested):
Seems Good to me, what remains i believe is solve issue for latency.

theanmolsharma (approved):
Approved, but let's also add some e2e tests

src/common/common.ts
```diff
@@ -147,3 +147,7 @@ export const varIntSize = (value: number): number => {
     else if (value <= 0xffffffff) return 5;
     else return 9;
 };
+
+export const delay = (ms: number): Promise<void> => {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+};
```
theanmolsharma, line 153: remove unused function

package-lock.json
theanmolsharma, line 1: I don't know what's wrong but when I run `npm install` it makes changes to this file. You probably forgot to commit ig?

theanmolsharma (changes requested):

seekersoftec:
Fixed all requested changes.

PR #48, opened 2024-09-30
Implement  Effective Reorg Handling Mechanism

aruokhai:
***Objective***
Reorgs Can be fatal to a blockchain indexer if not handled properly, this is because the during Reorgs,  blocks are replaced with other blocks, leading to inconsistent state of the indexer. This inconsistency must be handled effectively to ensure correctness of such indexer.

***Changes***
- Implemented Reorg Mechanism
- Improved `OperationState` handling  logic
-  Added Dynamic Cron Job Scheduling Mechanism
-  Added Verbosity Logging Option.

***Scope Of Change***
-The scope of change is critical because it affects  the core functioning of the indexer.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -3,7 +3,8 @@ db:
   synchronize: false
 app:
   port: <port>
-  schedulerInterval: <interval> #interval in seconds
+  verbose: <boolean>
+  debug: <boolean>
   network: <network>
   requestRetry:
     delay: <number> # delay in Milliseconds
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -3,6 +3,7 @@ db:
   synchronize: false
 app:
   port: <port>
+  schedulerInterval: <interval> #interval in seconds
   network: <network>
   requestRetry:
     delay: <number> # delay in Milliseconds
--- a/src/block-data-providers/base-block-data-provider.abstract.ts
+++ b/src/block-data-providers/base-block-data-provider.abstract.ts
@@ -5,44 +5,20 @@ import {
     TransactionInput,
     TransactionOutput,
 } from '@/indexer/indexer.service';
-import { TransactionsService } from '@/transactions/transactions.service';
-import { SchedulerRegistry } from '@nestjs/schedule';
-import { CronJob } from 'cron';
 import { ConfigService } from '@nestjs/config';
+import { BlockStateService } from '@/block-state/block-state.service';
+import { BlockState } from '@/block-state/block-state.entity';
 
-export interface BaseOperationState {
-    indexedBlockHeight: number;
-    blockCache: Record<number, string>;
-}
-
-export abstract class BaseBlockDataProvider<
-    OperationState extends BaseOperationState,
-> {
+export abstract class BaseBlockDataProvider<OperationState> {
     protected abstract readonly logger: Logger;
     protected abstract readonly operationStateKey: string;
-    protected cacheSize = 6;
-    protected readonly CRON_JOB_NAME = 'providerSync';
 
     protected constructor(
         protected readonly configService: ConfigService,
         private readonly indexerService: IndexerService,
         private readonly operationStateService: OperationStateService,
-        private readonly transactionService: TransactionsService,
-        private readonly schedulerRegistry: SchedulerRegistry,
-    ) {
-        const schedulerIntervalInSeconds = this.configService.get<string>(
-            'app.schedulerInterval',
-        );
-
-        const job = new CronJob(
-            `*/${schedulerIntervalInSeconds} * * * * *`,
-            () => this.sync(),
-        );
-        this.schedulerRegistry.addCronJob(this.CRON_JOB_NAME, job);
-        job.start();
-    }
-
-    abstract sync(): void;
+        protected readonly blockStateService: BlockStateService,
+    ) {}
 
     async indexTransaction(
         txid: string,
@@ -68,69 +44,38 @@ export abstract class BaseBlockDataProvider<
         )?.state;
     }
 
-    async setState(partialState: Partial<OperationState>): Promise<void> {
-        const oldState = (await this.getState()) || ({} as OperationState);
-
-        if (partialState.blockCache) {
-            const updatedBlockCache = {
-                ...oldState.blockCache,
-                ...partialState.blockCache,
-            };
-
-            if (this.cacheSize < Object.keys(updatedBlockCache).length) {
-                delete updatedBlockCache[oldState.indexedBlockHeight - 5];
-            }
-
-            partialState.blockCache = updatedBlockCache;
-        }
-
-        const newState = {
-            ...oldState,
-            ...partialState,
-        };
-
+    async setState(
+        state: OperationState,
+        blockState: BlockState,
+    ): Promise<void> {
         await this.operationStateService.setOperationState(
             this.operationStateKey,
-            newState,
+            state,
         );
+
+        await this.blockStateService.addBlockState(blockState);
     }
 
     abstract getBlockHash(height: number): Promise<string>;
 
     async traceReorg(): Promise<number> {
-        const { indexedBlockHeight, blockCache } = await this.getState();
-        let counter = indexedBlockHeight;
+        let state = await this.blockStateService.getCurrentBlockState();
 
-        if (Object.keys(blockCache).length === 0) {
-            return indexedBlockHeight;
+        if (!state) return null;
+
+        while (state) {
+            const fetchedBlockHash = await this.getBlockHash(state.blockHeight);
+
+            if (state.blockHash === fetchedBlockHash) return state.blockHeight;
+
+            await this.blockStateService.removeState(state);
+
+            this.logger.log(
+                `Reorg found at height: ${state.blockHeight}, Wrong hash: ${state.blockHash}, Correct hash: ${fetchedBlockHash}`,
+            );
+            state = await this.blockStateService.getCurrentBlockState();
         }
 
-        while (true) {
-            const storedBlockHash = blockCache[counter];
-
-            if (storedBlockHash === undefined) {
-                throw new Error('Reorgs levels deep');
-            }
-
-            const fetchedBlockHash = await this.getBlockHash(counter);
-
-            if (storedBlockHash === fetchedBlockHash) {
-                return counter;
-            }
-            console.log(
-                'reorg found at count: ',
-                counter,
-                ' and hash: ',
-                storedBlockHash,
-                ' ',
-                fetchedBlockHash,
-            );
-
-            await this.transactionService.deleteTransactionByBlockHash(
-                storedBlockHash,
-            );
-
-            --counter;
-        }
+        throw new Error('Cannot Reorgs, blockchain state exhausted');
     }
 }
--- a/src/block-data-providers/base-block-data-provider.abstract.ts
+++ b/src/block-data-providers/base-block-data-provider.abstract.ts
@@ -5,15 +5,44 @@ import {
     TransactionInput,
     TransactionOutput,
 } from '@/indexer/indexer.service';
+import { TransactionsService } from '@/transactions/transactions.service';
+import { SchedulerRegistry } from '@nestjs/schedule';
+import { CronJob } from 'cron';
+import { ConfigService } from '@nestjs/config';
 
-export abstract class BaseBlockDataProvider<OperationState> {
+export interface BaseOperationState {
+    indexedBlockHeight: number;
+    blockCache: Record<number, string>;
+}
+
+export abstract class BaseBlockDataProvider<
+    OperationState extends BaseOperationState,
+> {
     protected abstract readonly logger: Logger;
     protected abstract readonly operationStateKey: string;
+    protected cacheSize = 6;
+    protected readonly CRON_JOB_NAME = 'providerSync';
 
     protected constructor(
+        protected readonly configService: ConfigService,
         private readonly indexerService: IndexerService,
         private readonly operationStateService: OperationStateService,
-    ) {}
+        private readonly transactionService: TransactionsService,
+        private readonly schedulerRegistry: SchedulerRegistry,
+    ) {
+        const schedulerIntervalInSeconds = this.configService.get<string>(
+            'app.schedulerInterval',
+        );
+
+        const job = new CronJob(
+            `*/${schedulerIntervalInSeconds} * * * * *`,
+            () => this.sync(),
+        );
+        this.schedulerRegistry.addCronJob(this.CRON_JOB_NAME, job);
+        job.start();
+    }
+
+    abstract sync(): void;
 
     async indexTransaction(
         txid: string,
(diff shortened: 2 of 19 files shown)


config/config.yaml
```diff
@@ -3,6 +3,8 @@ db:
   synchronize: false
 app:
   port: <port>
+  schedulerInterval: <cron interval>
```
theanmolsharma, line 6: We should not make this configurable. I don't see any other indexer or similar service exposing internal configs like these. My philosophy is to be conservative, only expose what is required, and not bombard users with too many config options.

On the other hand, I agree that a 10s scheduling is too long so please change it to 1 second.

config/config.yaml
```diff
@@ -3,6 +3,8 @@ db:
...
+  verbose: <boolean>
```
theanmolsharma, line 7: also add option for `debug` level

theanmolsharma (commented):
Concept ACK
I like the approach although I don't love it.
I don't like the fact that we have to store the entire hash chain. I'd look at other RPCs like `getbestblockhash` or something.

More in-depth review coming soon.

config/config.yaml
aruokhai: If you are fine with it, i could make it `optional` which would also  align with your philosophy of being conservative. 

Also i think 10 seconds is quite fine :smile:

aruokhai:
> Concept ACK I like the approach although I don't love it. I don't like the fact that we have to store the entire hash chain. I'd look at other RPCs like `getbestblockhash` or something.
> 
> More in-depth review coming soon.

We Only store `6 blockhash` or whatever number is necessary, if reorg is levels deep, we result to error.

theanmolsharma:
> > Concept ACK I like the approach although I don't love it. I don't like the fact that we have to store the entire hash chain. I'd look at other RPCs like `getbestblockhash` or something.
> > More in-depth review coming soon.
> 
> We Only store `6 blockhash` or whatever number is necessary, if reorg is levels deep, we result to error.

This is not good, ideally we should be able to handle reorgs of any length. Reorgs greater than 6 are rare, but we should be prepared for the worst. Let's store the last 2016 blocks to be very conservative.

src/block-state/block-state.entity.ts
```diff
@@ -0,0 +1,10 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class BlockState {
+    @PrimaryColumn('integer')
+    indexedBlockHeight: number;
```
theanmolsharma, line 6:
```suggestion
    blockHeight: number;
```

src/block-state/block-state.entity.ts
```diff
@@ -0,0 +1,10 @@
...
+
+    @Column('text')
+    indexedBlockHash: string;
```
theanmolsharma, line 9:
```suggestion
    blockHash: string;
```

src/block-state/block-state.service.ts
```diff
@@ -0,0 +1,74 @@
...
+        const state = (
+            await this.blockStateRepository.find({
+                order: {
+                    indexedBlockHeight: 'DESC',
+                },
+                take: 1,
+            })
+        )[0];
+
+        return state;
```
theanmolsharma, line 27:
```suggestion
        return this.blockStateRepository.findOne({
            order: {
                indexedBlockHeight: 'DESC',
            },
        });
```

src/block-state/block-state.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    async addBlockState(state: BlockState): Promise<void> {
+        this.blockStateRepository.save(state);
```
theanmolsharma, line 31: missing await

src/block-state/block-state.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Remove and return the latest item in the state cache
+    async dequeue_operation_state(): Promise<BlockState | null> {
+        const latest_state = (
+            await this.blockStateRepository.find({
+                order: {
+                    indexedBlockHeight: 'DESC',
+                },
+                take: 1,
+            })
+        )[0];
```
theanmolsharma, line 45:
```suggestion
        const latestState = await this.blockStateRepository.findOne({
            order: {
                indexedBlockHeight: 'DESC',
            },
        });
```

src/block-state/block-state.service.ts
```diff
@@ -0,0 +1,74 @@
...
+                    indexedBlockHeight: 'ASC',
+                },
+                take: queueCount - this.cacheSize + 1,
+            });
+            await this.blockStateRepository.delete(
+                old_states.map((state) => state.indexedBlockHeight),
+            );
+        }
+    }
+}
```
theanmolsharma, line 74: > let's save the entire state ie don't trim to 2016 block

remove this

src/main.ts
```diff
@@ -10,6 +11,22 @@ async function bootstrap() {
...
+
+    const loggerLevels: LogLevel[] = ['error', 'warn', 'log'];
+
+    if (isVerbose) {
+        loggerLevels.push('verbose');
+    }
+
+    if (isDebug) {
+        loggerLevels.push('debug');
+    }
```
theanmolsharma, line 26: nit
```suggestion
    if (isVerbose) loggerLevels.push('verbose');
    if (isDebug) loggerLevels.push('debug');
```

src/main.ts
```diff
@@ -10,6 +11,22 @@ async function bootstrap() {
     const configService = app.get<ConfigService>(ConfigService);
     const port = configService.get<number>('app.port');
 
+    const isVerbose = configService.get<boolean>('app.verbose') ?? false;
+
+    const isDebug = configService.get<boolean>('app.debug') ?? false;
```
theanmolsharma, line 16: nit
```suggestion
    const isVerbose = configService.get<boolean>('app.verbose') ?? false;
    const isDebug = configService.get<boolean>('app.debug') ?? false;
```

src/block-data-providers/base-block-data-provider.abstract.ts
```diff
@@ -45,4 +49,30 @@ export abstract class BaseBlockDataProvider<OperationState> {
...
     }
+
+    abstract getBlockHash(height: number): Promise<string>;
+
+    async traceReorg(): Promise<number> {
+        let state = await this.blockStateService.getCurrentBlockState();
+
+        if (state == null) {
+            return null;
+        }
```
theanmolsharma, line 60: nit
```suggestion
        if (state == null) return null;
```

src/block-data-providers/base-block-data-provider.abstract.ts
```diff
@@ -45,4 +49,30 @@ export abstract class BaseBlockDataProvider<OperationState> {
...
+
+        while (true) {
```
theanmolsharma, line 62: I don't like `while true` loops because they could lead to nasty infinite loop bugs. We should replace it with a suitable condition.

src/block-data-providers/base-block-data-provider.abstract.ts
```diff
@@ -45,4 +49,30 @@ export abstract class BaseBlockDataProvider<OperationState> {
...
+            if (state === null) {
+                throw new Error('Reorgs levels deep');
+            }
```
theanmolsharma, line 65: nit
```suggestion
            if (state === null) throw new Error('Reorgs levels deep');
```

theanmolsharma: if we store the entire block chain, I don't think we'll ever run into this case

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -123,6 +133,11 @@ export class BitcoinCoreProvider
 
                 state.indexedBlockHeight = height;
                 await this.setState(state);
+
+                await this.blockStateService.addBlockState({
+                    indexedBlockHash: blockHash,
+                    indexedBlockHeight: height,
+                });
```
theanmolsharma, line 140: this should be done inside `setState`

theanmolsharma (changes requested):
nice work!!

theanmolsharma:
Here's a diff that addresses **most but not all** comments

```diff
diff --git a/src/block-data-providers/base-block-data-provider.abstract.ts b/src/block-data-providers/base-block-data-provider.abstract.ts
index 05b5e6c..e631d69 100644
--- a/src/block-data-providers/base-block-data-provider.abstract.ts
+++ b/src/block-data-providers/base-block-data-provider.abstract.ts
@@ -7,6 +7,7 @@ import {
 } from '@/indexer/indexer.service';
 import { ConfigService } from '@nestjs/config';
 import { BlockStateService } from '@/block-state/block-state.service';
+import { BlockState } from '@/block-state/block-state.entity';
 
 export abstract class BaseBlockDataProvider<OperationState> {
     protected abstract readonly logger: Logger;
@@ -43,11 +44,16 @@ export abstract class BaseBlockDataProvider<OperationState> {
         )?.state;
     }
 
-    async setState(state: OperationState): Promise<void> {
+    async setState(
+        state: OperationState,
+        blockState: BlockState,
+    ): Promise<void> {
         await this.operationStateService.setOperationState(
             this.operationStateKey,
             state,
         );
+
+        await this.blockStateService.addBlockState(blockState);
     }
 
     abstract getBlockHash(height: number): Promise<string>;
@@ -55,24 +61,16 @@ export abstract class BaseBlockDataProvider<OperationState> {
     async traceReorg(): Promise<number> {
         let state = await this.blockStateService.getCurrentBlockState();
 
-        if (state == null) {
-            return null;
-        }
+        if (state === null) return null;
 
         while (true) {
-            if (state === null) {
-                throw new Error('Reorgs levels deep');
-            }
+            if (state === null) throw new Error('Reorgs levels deep');
 
-            const fetchedBlockHash = await this.getBlockHash(
-                state.indexedBlockHeight,
-            );
+            const fetchedBlockHash = await this.getBlockHash(state.blockHeight);
 
-            if (state.indexedBlockHash === fetchedBlockHash) {
-                return state.indexedBlockHeight;
-            }
+            if (state.blockHash === fetchedBlockHash) return state.blockHeight;
 
-            state = await this.blockStateService.dequeue_operation_state();
+            state = await this.blockStateService.dequeueState();
         }
     }
 }
diff --git a/src/block-data-providers/bitcoin-core/provider.ts b/src/block-data-providers/bitcoin-core/provider.ts
index 5c8b9a4..b44bd16 100644
--- a/src/block-data-providers/bitcoin-core/provider.ts
+++ b/src/block-data-providers/bitcoin-core/provider.ts
@@ -73,15 +73,23 @@ export class BitcoinCoreProvider
             );
         } else {
             this.logger.log('No previous state found. Starting from scratch.');
-            const updatedState: BitcoinCoreOperationState = {
-                indexedBlockHeight:
-                    this.configService.get<BitcoinNetwork>('app.network') ===
-                    BitcoinNetwork.MAINNET
-                        ? TAPROOT_ACTIVATION_HEIGHT - 1
-                        : 0,
-            };
-
-            await this.setState(updatedState);
+
+            const blockHeight =
+                this.configService.get<BitcoinNetwork>('app.network') ===
+                BitcoinNetwork.MAINNET
+                    ? TAPROOT_ACTIVATION_HEIGHT - 1
+                    : 0;
+            const blockHash = await this.getBlockHash(blockHeight);
+
+            await this.setState(
+                {
+                    indexedBlockHeight: blockHeight,
+                },
+                {
+                    blockHash,
+                    blockHeight,
+                },
+            );
         }
     }
 
@@ -132,11 +140,9 @@ export class BitcoinCoreProvider
                 }
 
                 state.indexedBlockHeight = height;
-                await this.setState(state);
-
-                await this.blockStateService.addBlockState({
-                    indexedBlockHash: blockHash,
-                    indexedBlockHeight: height,
+                await this.setState(state, {
+                    blockHash: blockHash,
+                    blockHeight: height,
                 });
             }
         } finally {
diff --git a/src/block-data-providers/esplora/provider.ts b/src/block-data-providers/esplora/provider.ts
index 0403979..5dc36f5 100644
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -71,16 +71,25 @@ export class EsploraProvider
             );
         } else {
             this.logger.log('No previous state found. Starting from scratch.');
-            const updatedState: EsploraOperationState = {
-                currentBlockHeight: 0,
-                indexedBlockHeight:
-                    this.configService.get<BitcoinNetwork>('app.network') ===
-                    BitcoinNetwork.MAINNET
-                        ? TAPROOT_ACTIVATION_HEIGHT - 1
-                        : 0,
-                lastProcessedTxIndex: 0, // we dont take coinbase txn in account
-            };
-            await this.setState(updatedState);
+
+            const blockHeight =
+                this.configService.get<BitcoinNetwork>('app.network') ===
+                BitcoinNetwork.MAINNET
+                    ? TAPROOT_ACTIVATION_HEIGHT - 1
+                    : 0;
+            const blockHash = await this.getBlockHash(blockHeight);
+
+            await this.setState(
+                {
+                    currentBlockHeight: 0,
+                    indexedBlockHeight: blockHeight,
+                    lastProcessedTxIndex: 0, // we don't take coinbase txn into account
+                },
+                {
+                    blockHash,
+                    blockHeight,
+                },
+            );
         }
     }
 
@@ -162,11 +171,9 @@ export class EsploraProvider
 
                 state.indexedBlockHeight = height;
                 state.lastProcessedTxIndex = i + this.batchSize - 1;
-                await this.setState(state);
-
-                await this.blockStateService.addBlockState({
-                    indexedBlockHash: hash,
-                    indexedBlockHeight: height,
+                await this.setState(state, {
+                    blockHeight: height,
+                    blockHash: hash,
                 });
             } catch (error) {
                 this.logger.error(
diff --git a/src/block-state/block-state.entity.ts b/src/block-state/block-state.entity.ts
index 661700f..f31a7d7 100644
--- a/src/block-state/block-state.entity.ts
+++ b/src/block-state/block-state.entity.ts
@@ -3,8 +3,8 @@ import { Column, Entity, PrimaryColumn } from 'typeorm';
 @Entity()
 export class BlockState {
     @PrimaryColumn('integer')
-    indexedBlockHeight: number;
+    blockHeight: number;
 
     @Column('text')
-    indexedBlockHash: string;
+    blockHash: string;
 }
diff --git a/src/block-state/block-state.service.ts b/src/block-state/block-state.service.ts
index e1e53b8..0faa6d9 100644
--- a/src/block-state/block-state.service.ts
+++ b/src/block-state/block-state.service.ts
@@ -15,60 +15,32 @@ export class BlockStateService {
     ) {}
 
     async getCurrentBlockState(): Promise<BlockState> {
-        const state = (
-            await this.blockStateRepository.find({
-                order: {
-                    indexedBlockHeight: 'DESC',
-                },
-                take: 1,
-            })
-        )[0];
-
-        return state;
+        return this.blockStateRepository.findOne({
+            order: {
+                blockHeight: 'DESC',
+            },
+        });
     }
 
     async addBlockState(state: BlockState): Promise<void> {
-        this.blockStateRepository.save(state);
-        // Ensure the cache size does not exceed 2016
-        await this.trimState();
+        await this.blockStateRepository.save(state);
     }
 
     // Remove and return the latest item in the state cache
-    async dequeue_operation_state(): Promise<BlockState | null> {
-        const latest_state = (
-            await this.blockStateRepository.find({
-                order: {
-                    indexedBlockHeight: 'DESC',
-                },
-                take: 1,
-            })
-        )[0];
-
-        if (latest_state) {
-            await this.blockStateRepository.remove(latest_state);
+    async dequeueState(): Promise<BlockState | null> {
+        const latestState = await this.blockStateRepository.findOne({
+            order: {
+                blockHeight: 'DESC',
+            },
+        });
+
+        if (latestState) {
+            await this.blockStateRepository.remove(latestState);
             await this.transactionService.deleteTransactionByBlockHash(
-                latest_state.indexedBlockHash,
+                latestState.blockHash,
             );
-            return latest_state;
         }
 
-        return null;
-    }
-
-    private async trimState(): Promise<void> {
-        const queueCount = await this.blockStateRepository.count();
-
-        if (queueCount >= this.cacheSize) {
-            // Delete the oldest entries from the cache
-            const old_states = await this.blockStateRepository.find({
-                order: {
-                    indexedBlockHeight: 'ASC',
-                },
-                take: queueCount - this.cacheSize + 1,
-            });
-            await this.blockStateRepository.delete(
-                old_states.map((state) => state.indexedBlockHeight),
-            );
-        }
+        return latestState;
     }
 }
diff --git a/src/main.ts b/src/main.ts
index 68f5ccf..6488ce1 100644
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,18 +12,12 @@ async function bootstrap() {
     const port = configService.get<number>('app.port');
 
     const isVerbose = configService.get<boolean>('app.verbose') ?? false;
-
     const isDebug = configService.get<boolean>('app.debug') ?? false;
 
     const loggerLevels: LogLevel[] = ['error', 'warn', 'log'];
 
-    if (isVerbose) {
-        loggerLevels.push('verbose');
-    }
-
-    if (isDebug) {
-        loggerLevels.push('debug');
-    }
+    if (isVerbose) loggerLevels.push('verbose');
+    if (isDebug) loggerLevels.push('debug');
 
     app.useLogger(loggerLevels);
 
```

src/block-state/block-state.service.ts
aruokhai: does not work because selection statement is not added.

src/block-state/block-state.service.ts
```diff
@@ -0,0 +1,50 @@
...
+            await this.blockStateRepository.find({
+                order: {
+                    blockHeight: 'DESC',
+                },
+                take: 1,
+            })
+        )[0];
+
+        if (latestState) {
+            await this.blockStateRepository.remove({ ...latestState });
```
theanmolsharma, line 42:
```suggestion
            await this.blockStateRepository.delete(latestState.blockHeight);
```

theanmolsharma (commented):
Consider approved!! I need to run some reorg simulations to verify reorg handling. Will merge, once I'm confident that it works.

theanmolsharma (approved):
ACK 2327412828b22906f0faaf40d2412be53b438c3e

Excellent work!

PR #50, opened 2024-10-12
Add Makefile to Automate Local E2E Test Setup

aruokhai:
__Objective__
Introduce a Makefile to streamline the local setup and execution of End-to-End (E2E) tests, improving developer efficiency.

__Problem__
Running E2E tests requires specific processes to be set up before execution, and manually configuring these processes for local testing can be time-consuming and error-prone. This repetitive task reduces developer productivity and can lead to inconsistencies in the testing environment. Automating this setup with a Makefile will make local testing more efficient and reliable.

__Changes__
- Added a Makefile to automate the setup and execution of E2E tests locally.
- Added a README with clear instructions on how to run the tests using the new setup.

__Scope of Change__
This is a non-critical update, as it only affects the local environment for running E2E tests and does not impact production code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ lerna-debug.log*
 
 # misc
 /.silent-pay-indexer
+
+#e2e 
+/e2e/.logs
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -1,0 +1,55 @@
+# Variables
+DOCKER_COMPOSE_FILE = helpers/docker/docker-compose.yml
+NESTJS_SERVICE_PORT = 3000
+INDEXER_TEST_CMD = npm run test:e2e 
+MAX_RETRIES = 10 # Number of retries to check if the inddexer service is up
+RETRY_INTERVAL = 5 # Seconds between retries
+PM2_CMD = npx pm2
+NPM_CMD = npm
+APP_NAME = silent-indexer
+LOG_DIR = .logs
+INDEXER_OUTPUT_LOG = $(LOG_DIR)/$(APP_NAME)-out.log
+INDEXER_ERROR_LOG = $(LOG_DIR)/$(APP_NAME)-error.log
+
+# Targets
+.PHONY: all start-bitcoind start-indexer setup test clean 
+
+# Default target
+all: setup test
+	@$(MAKE) clean
+
+# Start Docker Compose services
+start-bitcoind:
+	@echo "Starting Docker Compose services..."
+	docker compose -f $(DOCKER_COMPOSE_FILE) up -d
+
+# Start the Indexer service
+start-indexer: 
+	@echo "Starting the Indexer service with PM2..."
+	$(PM2_CMD) start $(NPM_CMD) --name $(APP_NAME) --output $(INDEXER_OUTPUT_LOG) --error $(INDEXER_ERROR_LOG) -- run start:e2e 
+
+# Wait for the Indexer service to be available before running tests
+setup: start-bitcoind start-indexer
+	@echo "Waiting for Indexer service to be ready on port $(NESTJS_SERVICE_PORT)..."
+	@count=0; \
+	while ! nc -z localhost $(NESTJS_SERVICE_PORT) && [ $$count -lt $(MAX_RETRIES) ]; do \
+		count=$$((count + 1)); \
+		echo "Waiting for service... ($$count/$(MAX_RETRIES))"; \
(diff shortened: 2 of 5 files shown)


e2e/makefile
theanmolsharma, line 1: the standard naming convention is `Makefile`

e2e/README.md
```diff
@@ -0,0 +1,42 @@
...
+To install PM2 globally, follow these steps:
+
+1. **Open your terminal** (Command Prompt, PowerShell, or Terminal on macOS/Linux).
+
+2. **Run the following command** to install PM2 globally using npm:
+
+   ```bash
+   npm install -g pm2
+   ```
+
```
theanmolsharma, line 34: I don't like that the user needs to install a global dependency. We should add it as a dev dependency and do `npx pm2`

e2e/makefile
```diff
@@ -0,0 +1,54 @@
...
+
+# Start Docker Compose services
+up: down
+	@echo "Starting Docker Compose services..."
+	docker compose -f $(DOCKER_COMPOSE_FILE) up -d
+
+# Stop and remove Docker Compose services
+down:
+	@echo "Stopping and removing Docker Compose services..."
+	docker compose -f $(DOCKER_COMPOSE_FILE) down
```
theanmolsharma, line 25:
```suggestion
	docker compose -f $(DOCKER_COMPOSE_FILE) down -v
```

e2e/makefile
```diff
@@ -0,0 +1,54 @@
...
+	fi
+
+# Run the end-to-end tests
+test: up start-service wait-for-service
+	@echo "Running end-to-end tests..."
+	$(NESTJS_TEST_CMD)
+
+# Clean up
+clean: down
+	@echo "Clean up completed."
```
theanmolsharma, line 54: redundant with `down`, please remove

theanmolsharma (changes requested):
1. I want you to look into if we absolutely need `pm2`, I think we should be able to achieve something similar without it.
2. We need a command or something to view the indexer logs during/after the run. We should be able to use `tail` command to achieve this.

e2e/makefile
aruokhai: The `-v` option is meant to remove named volumes, which we do not use.

aruokhai:
The problem with using '&' to start background tasks is that they must be terminated manually. In contrast, PM2 simplifies this process by assigning an identifier for easy management and control

e2e/makefile
theanmolsharma: It also removes anonymous volumes which are used in the image.

aruokhai:
@theanmolsharma, Using the tail command has also been challenging due to the issues with background PIDs. As an alternative, I opted to use output log files.

.gitignore
```diff
@@ -35,4 +35,7 @@ lerna-debug.log*
 !.vscode/extensions.json
 
 # misc
-/.silent-pay-indexer
\ No newline at end of file
+/.silent-pay-indexer
+
+#e2e 
+/.logs
```
theanmolsharma, line 41: not required because we already have `*.log` on L7

theanmolsharma: If you want to keep it still, it should be `/e2e/.logs/` not `/.logs/`

theanmolsharma (changes requested):

.gitignore
aruokhai: yep, it was meant to be that :smiling_face_with_tear:

e2e/Makefile
```diff
@@ -0,0 +1,55 @@
...
+	done; \
+	if ! nc -z localhost $(NESTJS_SERVICE_PORT); then \
+		echo "Error: Service not available after $(MAX_RETRIES) retries, exiting."; \
+		exit 1; \
+	fi
+
+# Run the end-to-end tests
+test: 
+	@echo "Running end-to-end tests..."
+	$(INDEXER_TEST_CMD)
```
theanmolsharma, line 48: This should be something like
```suggestion
	$(INDEXER_TEST_CMD) || echo "Warning: E2E tests failed, but continuing..."
```

motivation: I ran `make` locally, and the tests failed, but the cleanup didn't happen because the process exits when an error happens (similar to shell script)

theanmolsharma (changes requested):

theanmolsharma (approved):
ACK 2e86330e8513004a84bb3dfc7afbdd08487c50b0

PR #54, opened 2024-11-05
Add Dockerfile For Deployment

### LINKED ISSUE #53: Create A Dockerfile For Silent Indexer
The Silent Payment Indexer project currently lacks a standardized deployment setup

 The addition of a Dockerfile would afford us to a initialization to begin the development of the Development Pipeline.

aruokhai:
Added the needed Dockerfile necessary for Deployment

closes #53


config/config.template.yaml
theanmolsharma, line 1: rename to `example.config.yaml` and provide standard config values

migrate.js
theanmolsharma, line 1: We don't need this. Please use typeorm for migration generation and management

theanmolsharma: You can add a npm script like `"migration:run": "typeorm-ts-node-commonjs migration:run -d migrations/data-source.ts -t each",`

migrations/002_create_operation_state_table.sql
theanmolsharma, line 1: I don't see a good reason for having two migrations files. Please merge into just one migration.

dockerfile
```diff
@@ -0,0 +1,44 @@
+# Stage 1: Build the NestJS application
+FROM node:18-alpine AS builder
```
theanmolsharma, line 2: we should use use v22 as it is current LTS version

dockerfile
theanmolsharma, line 1: standard naming convention is `Dockerfile`

dockerfile
```diff
@@ -0,0 +1,44 @@
...
+WORKDIR /app
+
+# Copy package.json and package-lock.json for efficient layer caching
+COPY package*.json ./
+
+# Install dependencies
+RUN npm install
+
+# Copy the rest of the application code
+COPY . .
```
theanmolsharma, line 14: instead of copying everything we should only copy what is required. This includes:
- `tsconfig.json`
- `tsconfig.build.json`
- `nest-cli.json`
- `webpack.config.js`
- `src`
- `config`
- `migrations`
- `package-lock.json`
- `package.json`

theanmolsharma (changes requested):
Let's remove the migration code from this PR. I'll make a PR for it.

dockerfile
theanmolsharma, line 1: I don't think the comments you've added here are useful.
eg the code is pretty self explanatory 
```
# Set the working directory in the container
WORKDIR /app
```

config/config.template.yaml
aruokhai: It is going to affect  the usage of ` envsubst ` util command, we can put them as comments

dockerfile
theanmolsharma: Even though it's done in the `build` section, I'm pretty sure that it's the standard practice to just copy the required files.

config/config.template.yaml
```diff
@@ -0,0 +1,42 @@
...
+  rpcHost: ${RPC_HOST}
+  rpcPass: ${RPC_PASS}
+  rpcUser: ${RPC_USER}
+  rpcPort: ${RPC_PORT}
+
+
+### Example ####
+# db:
+#   path: ':memory:'
+#   synchronize: true
```
theanmolsharma, line 27: this should be false

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /app
+
+RUN apk add --no-cache gettext
+
+#Copy the compiled application from the build stage
+COPY --from=builder /app/dist ./dist
+COPY --from=builder /app/migrations ./migrations
+
+#Copy package.json and install only production dependencies
+COPY package.json ./
```
theanmolsharma, line 30: why are we not copying the lock file here?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN npm install --only=production
+
+# Set environment variables for SQLite path and app port
+ENV SQLITE_PATH="/app/.silent-pay-indexer/db/database.sqlite"
```
theanmolsharma, line 34:
```suggestion
ENV SQLITE_PATH="/data/.silent-pay-indexer/db/database.sqlite"
```

config/config.template.yaml
theanmolsharma: We don't need `envsubst`, the config values provided in the yaml files are overwritten with env variables once the app starts. [[ref](https://github.com/Bitshala-Incubator/silent-pay-indexer/blob/2327412828b22906f0faaf40d2412be53b438c3e/src/configuration.ts#L47)]

Dockerfile
aruokhai: It is not copied because it is not needed to run the application.

config/example.config.yaml
```diff
@@ -0,0 +1,18 @@
+db:
+  path: ':memory:'
```
theanmolsharma, line 2: this should be `~/.silent-pay-indexer/db/database.sqlite`

Dockerfile
theanmolsharma: yeah but it is required to download the dependencies right

config/example.config.yaml
```diff
@@ -0,0 +1,18 @@
...
+  synchronize: false
+app:
+  port: 3000
+  verbose: false
+  debug: true
+  network: regtest
+  requestRetry:
+    delay: 500       # delay in Milliseconds
+    count: 1
```
theanmolsharma, line 11: lets set it to 3

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+#Copy the compiled application from the build stage
+COPY --from=builder /app/dist ./dist
+COPY --from=builder /app/migrations ./migrations
+
+#Copy package.json and install only production dependencies
+COPY package.json ./
+RUN npm install --only=production
+
+# Set environment variables for SQLite path and app port
+ENV DB_PATH="/app/data/.silent-pay-indexer/db/database.sqlite"
```
theanmolsharma, line 32:
```suggestion
ENV DB_PATH="/app/data/database.sqlite"
```

theanmolsharma (changes requested):

theanmolsharma (approved):

PR #55, opened 2024-11-07
feat: add scripts to generate and run migrations using typeorm

theanmolsharma:


migrations/data-source.ts
```diff
@@ -0,0 +1,22 @@
...
+    return value;
+}
+
+async function configureDataSource(): Promise<DataSource> {
+    return new DataSource({
+        database: getEnvVariable('DB_PATH'),
+        type: 'sqlite',
+        synchronize: false,
+        logging: false,
+        entities: ['src/**/**.entity.ts'],
```
aruokhai, line 17: I don't think this is necessary

theanmolsharma: I'm pretty sure that it's necessary, if we don't then how will Typeorm know which files to use for entity generation?

PR #59, opened 2024-11-16
fix: make indexing transactional

chaitika:


src/db-transaction/db-transaction.service.ts
```diff
@@ -0,0 +1,40 @@
...
+            this.queryRunnerSet.delete(queryRunner);
+        }
+    }
+
+    async onModuleDestroy(): Promise<void> {
+        for (const queryRunner of this.queryRunnerSet) {
+            await queryRunner.rollbackTransaction();
+        }
+    }
+}
```
aruokhai, line 40: I need your help understanding the necessity of this function. To the best of my knowledge, the `finally` clause in the `execute` function ensures that  created resources are released . In what scenario would such resources not be released and `rollback` needed

src/db-transaction/db-transaction.service.ts
```diff
@@ -0,0 +1,40 @@
+import { Injectable, OnModuleDestroy } from '@nestjs/common';
+import { DataSource, EntityManager, QueryRunner } from 'typeorm';
+import { IsolationLevel } from 'typeorm/driver/types/IsolationLevel';
+
+@Injectable()
+export class DbTransactionService implements OnModuleDestroy {
+    private readonly queryRunnerSet: Set<QueryRunner>;
```
aruokhai, line 7: In the advent that `onModuleDestroy` function is not needed as suggested in  a comment below , i believe this variable existence would be futile

src/db-transaction/db-transaction.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    constructor(private readonly dataSource: DataSource) {
+        this.queryRunnerSet = new Set();
```
aruokhai, line 9: same logic also applies here

src/db-transaction/db-transaction.service.ts
```diff
@@ -0,0 +1,40 @@
...
+        } catch (err) {
+            // since we have errors, rollback the changes we made
+            await queryRunner.rollbackTransaction();
+            throw err;
+        } finally {
+            // we need to release a queryRunner which was manually instantiated
+            await queryRunner.release();
+            this.queryRunnerSet.delete(queryRunner);
+        }
+    }
```
aruokhai, line 33: If all suggested is true, such function needs no separate folder, it could be chunked in the common utils folder

src/block-data-providers/esplora/provider.ts
```diff
@@ -144,36 +149,45 @@ export class EsploraProvider
...
+                    state.indexedBlockHeight = height;
+                    state.lastProcessedTxIndex = i + this.batchSize - 1;
+                    await this.setState(
+                        state,
+                        {
+                            blockHeight: height,
+                            blockHash: hash,
+                        },
+                        manager,
+                    );
```
theanmolsharma, line 190: We are doing computation, even making a network call inside the transaction. This is not ideal and needs to be updated either in this PR or a followup PR.

So all the computation should happen outside of the transaction, and in the transaction we just update the entities atomically

src/block-data-providers/bitcoin-core/provider.ts
```diff
@@ -127,22 +132,29 @@ export class BitcoinCoreProvider
...
+                    }
+
+                    state.indexedBlockHeight = height;
+                    await this.setState(
+                        state,
+                        {
+                            blockHash: blockHash,
+                            blockHeight: height,
+                        },
+                        manager,
```
theanmolsharma, line 156: same here, computation should happen outside of the transaction

src/db-transaction/db-transaction.service.ts
aruokhai: Okay

src/block-data-providers/esplora/provider.ts
chaitika: okay, we should do it in a followup PR because the method signatures will change again

theanmolsharma (approved):

PR #64, opened 2024-12-13
Add Support for P2PKH, P2TR, P2SH-P2WPKH Address Types 

notTanveer:
This PR adds support for testing P2PKH, P2TR,  P2SH-P2WPKH address types in the indexer, ensuring silent blocks are correctly processed and retrieved for these transaction types. The test suite has been refactored improving maintainability and extending compatibility with diverse Bitcoin address types.


e2e/helpers/wallet.helper.ts
```diff
@@ -53,12 +65,18 @@ export class WalletHelper {
         return this.bitcoinRPCUtil.mineToAddress(numOfBlocks, walletAddress);
     }
 
-    async addFundToUTXO(payment: Payment, amount): Promise<UTXO> {
+    async addFundToUTXO(
+        payment: Payment,
+        amount: number,
+        addressType: AddressType,
+        index: number,
```
theanmolsharma, line 72: you are passing this to the function and just returning it. If you don't need to use it in the function then why pass it as an argument?

e2e/helpers/wallet.helper.ts
```diff
@@ -157,4 +198,84 @@ export class WalletHelper {
...
+        case AddressType.P2PKH:
+            return `m/44'/0'/0'/0/${index}`;
+        case AddressType.P2SH_P2WPKH:
+            return `m/49'/0'/0'/0/${index}`;
+        case AddressType.P2WPKH:
+            return `m/84'/0'/0'/0/${index}`;
+        case AddressType.P2TR:
+            return `m/86'/0'/0'/0/${index}`;
+        default:
+            throw new Error('Unsupported address type');
```
theanmolsharma, line 263: for testnet we should `1'` as coin type
```suggestion
        case AddressType.P2PKH:
            return `m/44'/1'/0'/0/${index}`;
        case AddressType.P2SH_P2WPKH:
            return `m/49'/1'/0'/0/${index}`;
        case AddressType.P2WPKH:
            return `m/84'/1'/0'/0/${index}`;
        case AddressType.P2TR:
            return `m/86'/1'/0'/0/${index}`;
        default:
            throw new Error('Unsupported address type');
```

e2e/helpers/wallet.helper.ts
```diff
@@ -157,4 +198,84 @@ export class WalletHelper {
...
+    }
+}
+
+function createTaprootKeyPair(
+    keyPair: BIP32Interface,
+    network = networks.regtest,
```
theanmolsharma, line 269: is there any scenario where network won't be regtest?
If not, then this shouldn't be tan argument

e2e/helpers/wallet.helper.ts
```diff
@@ -157,4 +198,84 @@ export class WalletHelper {
 
         return { transaction, txid, blockhash };
     }
+
+    private createInputFromUtxo(utxo: UTXO, keyPair: BIP32Interface): any {
+        const input: any = {
```
theanmolsharma, line 203:
```suggestion
    private createInputFromUtxo(
        utxo: UTXO,
        keyPair: BIP32Interface,
    ): PsbtInputExtended {
        const input: PsbtInputExtended = {
```

and add this somewhere on top

```ts
export interface PsbtInputExtended extends PsbtInput, TransactionInput {}
```

e2e/indexer.e2e-spec.ts
```diff
@@ -21,46 +21,64 @@ describe('Indexer', () => {
...
+        AddressType.P2TR,
+        AddressType.P2PKH,
+        AddressType.P2SH_P2WPKH,
+    ];
 
-        for (const output of p2wkhOutputs) {
-            const utxo = await walletHelper.addFundToUTXO(output, 1);
-            utxos.push(utxo);
-        }
+    it.each(addressTypes)(
```
theanmolsharma, line 31: This should achieve the same result
```suggestion
    it.each(Object.values<AddressType>(AddressType))(
```

e2e/indexer.e2e-spec.ts
```diff
@@ -21,46 +21,64 @@ describe('Indexer', () => {
...
-            },
-        );
+            const silentBlock = new SilentBlocksService(
+                {} as any,
+                {} as any,
+            ).encodeSilentBlock([transformedTransaction]);
 
-        expect(response.data).toEqual(silentBlock);
-    });
+            await new Promise((resolve) => setTimeout(resolve, 15000));
```
theanmolsharma, line 73: No action required for this PR
but we should be able to get rid of this once we start listening for WS updates

theanmolsharma (changes requested):
Additional changes:
1. please rename `blockhash` to `blockHash` everywhere

Nice work. Will merge once you address all comments.

e2e/helpers/wallet.helper.ts
notTanveer: this is used for 3 main reasons:
1. providing complete UTXO information
2. maintaining context for downstream processes
3. consistency and clarity

e2e/helpers/wallet.helper.ts
theanmolsharma:
```ts
import { PsbtInput } from 'bip174/src/lib/interfaces';
import { TransactionInput } from 'bitcoinjs-lib/src/psbt';

export interface PsbtInputExtended extends PsbtInput, TransactionInput {}
```

PR #68, opened 2025-01-08
Optional Flag filterSpent , Needed To Exclude Spent Transaction 

aruokhai:
### Objective
Transactions often contain outputs that are entirely spent shortly after being created. Excluding such transactions from Silent Blocks can lead to significant bandwidth and processing cost savings.

The goal of this task is to add an optional `filterSpent `flag to the **fetchSilentBlockByHash**() function. When this flag is set, transactions with fully spent outputs should be excluded from the Silent Blocks.

### Scope Of Change
Critical


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
notTanveer, line 1: i think you committed this by mistake. kindly revert

theanmolsharma (changes requested):
Really awesome work!! Keep the PRs coming!!

src/transactions/transaction-output.entity.ts
```diff
@@ -0,0 +1,26 @@
+import { Transaction } from '@/transactions/transaction.entity';
+import { Entity, Column, PrimaryGeneratedColumn, ManyToOne } from 'typeorm';
+
+@Entity()
+export class TransactionOutput {
+    @PrimaryGeneratedColumn()
+    id: number; // Auto-generated primary key
+
+    @Column({ type: 'text', nullable: false })
```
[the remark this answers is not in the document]
aruokhai: I don't think there is defined length of `scriptPubKey`.

src/transactions/transaction-output.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    @Column({ type: 'integer', nullable: false })
+    value: number; // Value of the output
+
+    @Column({ type: 'boolean', nullable: false, default: false })
+    isSpent: boolean;
+
+    @ManyToOne(() => Transaction, (transaction) => transaction.outputs, {
+        eager: true,
+        onDelete: 'CASCADE',
```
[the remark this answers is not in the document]
aruokhai: Verified the assumption : https://orkhan.gitbook.io/typeorm/docs/relations#relation-options

src/transactions/transaction.entity.ts
```diff
@@ -20,9 +15,10 @@ export class Transaction {
     @Column({ type: 'text', nullable: false })
```
notTanveer, line 15: Update the transaction class to use a VARCHAR data type and apply a length constraint.

src/indexer/indexer.service.ts
```diff
@@ -30,42 +28,67 @@ export class IndexerService {
...
+
+        // Mark each output as spent
+        const spentOutputs = outputs.map((output) => ({
+            ...output,
+            isSpent: true,
+        }));
+
+        await manager.save(TransactionOutputEntity, spentOutputs);
+    }
+
```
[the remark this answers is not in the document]
aruokhai: For the first opinion, How do you suggest i update per block, as insertion is done per transaction ?

src/transactions/transaction.entity.ts
aruokhai: this can be a seperate PR

migrations/1736332353003-migrations.ts
theanmolsharma, line 1: This is still wrong, you are dropping the column before migrating the data.
Also you should do everything in the same transaction so that it's atomic

src/transactions/transaction-output.entity.ts
```diff
@@ -0,0 +1,36 @@
+import { Transaction } from '@/transactions/transaction.entity';
+import { Entity, Column, PrimaryColumn, ManyToOne, JoinColumn } from 'typeorm';
+import { TransactionOutput as Output } from '@/indexer/indexer.service';
+
+@Entity()
+export class TransactionOutput {
+    static from(output: Output, index: number): TransactionOutput {
```
theanmolsharma, line 7:
```suggestion
    static fromOutput(output: Output, index: number): TransactionOutput {
```

theanmolsharma (changes requested):

src/transactions/transaction-output.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+        const newOutput = new TransactionOutput();
+        newOutput.pubKey = output.scriptPubKey.substring(4);
+        newOutput.value = output.value;
+        newOutput.vout = index;
+
+        return newOutput;
+    }
+
+    @PrimaryColumn({ type: 'text', nullable: false })
```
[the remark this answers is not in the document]
aruokhai: VARCHAR doesn't exist in Sqlite as a datatype

migrations/1736332353003-migrations.ts
aruokhai: I believe that removing the `isSpent` column early should not cause any issues, as it was not being used before the migration.

theanmolsharma:
Closing in favour of #74

theanmolsharma:
Closing in favour of #74

PR #71, opened 2025-04-14
fix: vulnerable deps

notTanveer:
## Changes
- Updated various dependency packages to their latest secure versions
- Fixed security vulnerabilities in transitive dependencies
-  Updated core dependencies while maintaining compatibility with existing code


theanmolsharma (commented):
I don’t believe a major version upgrade is necessary to resolve the dependency issues. Please go ahead and close this PR—I’ll handle the upgrade separately.

PR #75, opened 2025-05-05
Implement rate limiting across major services

notTanveer:
This PR introduces rate limiting to the application to enhance security and prevent abuse from excessive requests.

## Changes:
Imported ThrottlerModule into app.module.ts.
Set up a global default rate limit (5 requests per second).
Registered ThrottlerGuard globally using APP_GUARD to protect all routes by default.

## Benefits:
Protects the application against brute-force attacks.
Prevents individual users or IP addresses from overwhelming the server with too many requests.
Improves overall application stability and availability.
## Test
image
image

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -10,6 +10,8 @@ import { OperationStateModule } from '@/operation-state/operation-state.module';
 import { ScheduleModule } from '@nestjs/schedule';
 import { BlockProviderModule } from '@/block-data-providers/block-provider.module';
 import { EventEmitterModule } from '@nestjs/event-emitter';
+import { ThrottlerModule, ThrottlerGuard } from '@nestjs/throttler';
+import { APP_GUARD } from '@nestjs/core';
 
 @Module({
     imports: [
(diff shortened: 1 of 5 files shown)


src/silent-blocks/silent-blocks.controller.ts
```diff
@@ -1,7 +1,9 @@
 import { Controller, Get, Param, Res } from '@nestjs/common';
 import { Response } from 'express';
 import { SilentBlocksService } from '@/silent-blocks/silent-blocks.service';
+import { Throttle } from '@nestjs/throttler';
 
+@Throttle({ default: { limit: 10, ttl: 60000 } })
```
theanmolsharma, line 6: I don't think this is required once you've added global request throttling in `app.module.ts`

src/app.module.ts
```diff
@@ -34,8 +36,20 @@ import { EventEmitterModule } from '@nestjs/event-emitter';
         SilentBlocksModule,
         OperationStateModule,
         BlockProviderModule,
+        ThrottlerModule.forRoot([
+            {
+                ttl: 60000,
+                limit: 10,
```
theanmolsharma, line 42: What's TTL?
Does this mean that I can make only 10 requests in 60 seconds?
We should make this configurable.

theanmolsharma (commented):
LGTM.
Maybe break the change into two commits.
- chore(deps): add @nestjs/throttler
- feat: add request throttling

.gitignore
```diff
@@ -34,7 +34,7 @@ lerna-debug.log*
 !.vscode/extensions.json
 
 # misc
-/.silent-pay-indexer
+~/.silent-pay-indexer
```
theanmolsharma, line 37: Why?

notTanveer: apparently this is the correct path to the db.

theanmolsharma: The DB path is set in the config file, so it really depends on your local setup. Also, Git shouldn't be tracking anything outside the repo — `~` points to your home directory, which isn't part of the project and shouldn't be committed anyway. I'm not sure why this change is needed?

notTanveer: i guess it's my local set up then, reverting changes.

theanmolsharma (approved):

PR #79, opened 2025-05-12
refactor: make cache TTL configurable

chaitika:


theanmolsharma (changes requested):
Please update `src/configuration.model.ts` as well.

PR #108, opened 2026-06-28
perf: optimize range API with streaming & prefetch

notTanveer:
## Summary

- **Streaming range endpoint**: Replace full-buffer approach with chunked byte streaming and `Cache-Control` headers.
- **Block prefetch**: Prefetch the next block in both Bitcoin Core and Esplora providers to hide RPC/HTTP latency during sequential scans
- **Binary precomputation**: Precompute and store silent block data in binary format at index time, eliminating per-request serialization
- **WebSocket streaming sync**: Add `filterSpent` support and batched binary framing to the WS gateway for efficient wallet sync
- **Keepalive**: ping/pong handler so Cloudflare does not drop an idle sync socket
- **Empty frames**: skip heights with no silent payments. Over 99% of mainnet blocks are empty, so the stream was mostly no-op frames

### Benchmark Results (tested with shroud)

| Architecture | 50-Block Sync Time | Client Throughput | Payload per 1,000 Blocks |
| :--- | :--- | :--- | :--- |
| **V1: REST JSON (`filterSpent=true`)** | ~6.00s | ~2,500 tx/sec | ~20 MB |
| **V2: REST Binary (Unfiltered)** | ~1.50s | ~16,700 tx/sec | ~58 MB |
| **V3: REST Binary (`filterSpent=true`)** | ~2.00s | ~9,500 tx/sec * | ~12 MB |
| **V4: WebSocket Binary Streaming (Current)** | **< 0.50s** | **~20,000 tx/sec** | **~12 MB** |

### Key Wins
1. **Network Pipelining:** WebSocket streaming eliminates the HTTP overhead and constant polling latency.
2. **Maximum Wasm Efficiency:** The Rust WebAssembly engine is now parsing heavily filtered UTXOs at a blistering **20,000 transactions per second**. Batch scanning consistently finishes in `~570ms` for massive 1.5MB chunks.
3. **Bandwidth Reduction:** Implementing `filterSpent=true` strictly on the binary endpoint reduced sync payload size by **~80%**, avoiding the Base64/JSON serialization tax entirely.

image

### Indexer

* **Reset `lastProcessedTxIndex` after each block.** It was never reset when a block finished, causing every subsequent block to start past the end of its transaction list. Nothing was saved and the indexed height never advanced, stalling sync.
* **Persist resumed blocks correctly.** When block processing resumed mid-way, the resulting silent block could be truncated. The block is now read back from committed storage and written in its own batch.
* **Handle failed look-ahead writes.** If block `N` failed to write, the promise for block `N+1` could become orphaned and its late rejection could terminate the process.
* **Emit `INDEXED_BLOCK_EVENT` once per block.** Previously, the event fired once per transaction batch, causing a 900-tx block to broadcast 9 times, with 8 broadcasts containing partial data.
* **Sort silent blocks by txid.** Payload ordering previously depended on the producer, meaning the same block height could produce different payloads despite the bytes being served as immutable data.
* **Commit block state for every block.** A coinbase-only block runs zero transaction batches, so its height was never marked done — while the silent block was still saved and broadcast. Every restart reprocessed and re-broadcast that height. State now commits in the same batch as the blob, so the two cannot diverge.
* **Cover the whole block in the error handler.** The readback and silent block commit sat outside the `try`/`catch`, so a failure there threw without naming the height.

### HTTP

* **Validate block heights at the edge.** `/range` validated the requested span but not the height itself. Heights above `uint32` reached `writeUInt32BE` and resulted in a 500. `transactions/range` had the same issue.
* **Prevent truncated responses from being cached.** A mid-stream throw could result in a short body being returned as a successful `200` with `immutable, max-age=1y`, causing clients and CDNs to cache truncated ranges for a year. The socket is now destroyed instead.
* **Fix backpressure handling.** The backpressure helper could attach its `close` listener after the event had already fired and hang indefinitely. It also had no timeout. This is now replaced with `pipeline()` and a 60-second stall timeout, matching the WebSocket path.
* **Clamp `endHeight` to the chain tip.** Empty heights are now skipped, preventing an unindexed height from being treated the same as an empty height and causing clients to advance their cursor past it.
* **Fix caching and validation for `height/:height`.** The endpoint no longer serves immutable cache headers for the current tip. It now requires 6 confirmations, matching `/range`, and also uses `ParseIntPipe` for height validation.
* **404 an unindexed height on `height/:height`.** A height above the tip returned the same 2 bytes as a genuinely empty block, so clients read "not indexed yet" as "no payments here". `/range` already clamped for this reason.
* **Rate-limit `/range`.** This is the most expensive HTTP endpoint and previously had no rate limit.

### WebSocket

* **Allow only one sync per socket.** Concurrent sync messages on the same socket could mutate the shared syncing set at the same time and corrupt its state.
* **Prevent broadcast frames from corrupting sync parsing.** Tip broadcasts are bare blobs, while sync frames use `height | len | blob`. A broadcast arriving mid-sync could therefore desynchronize the client's frame parser. Syncing clients are now skipped for broadcasts.
* **Fix `from` parsing.** Passing `null` or `""` through `Number()` could unintentionally trigger a full sync from genesis.
* **Fix `filterSpent` parsing.** Using `??` preserved the string `"false"` as truthy, causing spent outputs to be filtered even when the client explicitly requested otherwise.
* **Bound drain waits.** The drain loop previously polled every 5ms with no upper bound, allowing a half-open connection to spin for roughly 15 minutes. It now backs off to 100ms and terminates clients that make no draining progress for 60 seconds.
* **Stop dropping blocks indexed during a sync.** `to` is fixed at sync start and broadcasts skip syncing clients, so a block indexed mid-sync was excluded from both — dropped with no signal. Sync now re-checks the tip after each pass and streams the delta, capped at 3 rounds so a client syncing during IBD cannot hold the socket. Hitting the cap acks with `tip > to`, which tells the client to sync again.

### Backfill

* **Prevent stale backfill writes.** Backfill previously encoded 100 heights, yielded, and then wrote stale bytes over data that may have been committed during that window. It now performs read → encode → commit per height, with the yield happening after the commit.

### Tests

**108 tests pass.**

## Not in this PR

* **Partial-block recovery:** If the process crashes mid-block, `lastProcessedTxIndex` can survive and cause the next block to skip its first `N` transactions. The proper fix is to avoid advancing `indexedBlockHeight` for a partially processed block, which requires a larger change than this PR.
* **WebSocket sync limits:** WS sync currently has no range cap or rate limit unlike REST. A single socket carrying the entire backfill is intentional.
* **`hash/:hash` tip check:** `height/:height` now 404s an unindexed height, but a hash carries no ordering and there is no block-hash → height index, so an unknown hash and a coinbase-only block both scan to zero rows. Distinguishing them needs a new index, not a guard.
* **`filterSpent` defaults:** WebSocket defaults to `true` while HTTP defaults to `false`. The ~12 MB payload reduction depends on the WebSocket default.

### DIFF
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -128,11 +128,19 @@ export class EsploraProvider
             let height =
                 ((await this.traceReorg()) ?? state.indexedBlockHeight) + 1;
 
+            let nextBlockHash: Promise<string> | null =
+                this.getBlockHash(height);
+
             for (height; height <= tipHeight; height++) {
-                const blockHash = await this.getBlockHash(height);
+                const blockHash = await nextBlockHash;
                 this.logger.log(
                     `Processing block at height ${height}, hash ${blockHash}`,
                 );
+
+                nextBlockHash =
+                    height + 1 <= tipHeight
+                        ? this.getBlockHash(height + 1)
+                        : null;
 
                 await this.processBlock(height, blockHash);
             }
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -229,8 +229,6 @@ export class EsploraProvider
                         batch,
                     );
                 });
-
-                this.eventEmitter.emit(INDEXED_BLOCK_EVENT, height);
             } catch (error) {
                 this.logger.error(
                     `Error processing transactions in block at height ${height}, hash ${hash}: ${error.message}`,
@@ -253,6 +251,11 @@ export class EsploraProvider
                 encodeSilentBlock(blockTxData),
             );
         });
+
+        // Emitted once per block, after the blob is stored: a per-batch emit
+        // broadcast the same height repeatedly, each time re-encoding whatever
+        // subset of the block had been committed so far.
+        this.eventEmitter.emit(INDEXED_BLOCK_EVENT, height);
     }
 
     private async getTipHeight(): Promise<number> {
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -17,6 +17,8 @@ import { DbTransactionService } from '@/db-transaction/db-transaction.service';
 import { EventEmitter2 } from '@nestjs/event-emitter';
 import { INDEXED_BLOCK_EVENT } from '@/common/events';
 import { StorageService } from '@/storage/storage.service';
+import { encodeSilentBlock } from '@/silent-blocks/silent-block-encoder';
+import { TransactionData } from '@/storage/interfaces';
 
 @Injectable()
 export class EsploraProvider
@@ -142,6 +144,7 @@ export class EsploraProvider
     private async processBlock(height: number, hash: string) {
         const state = await this.getState();
         const txids = await this.getTxidsForBlock(hash);
+        const allBlockTxData: TransactionData[] = [];
 
         for (
             let i = state.lastProcessedTxIndex + 1;
@@ -152,6 +155,7 @@ export class EsploraProvider
                 i,
                 Math.min(i + this.batchSize, txids.length),
             );
+            const isLastBatch = i + this.batchSize >= txids.length;
 
             try {
                 await this.dbTransactionService.execute(async (batch) => {
@@ -182,7 +186,7 @@ export class EsploraProvider
                                 value: output.value,
                             }));
 
-                            const saved = await this.indexTransaction(
+                            const result = await this.indexTransaction(
                                 txid,
                                 vin,
                                 vout,
@@ -192,9 +196,11 @@ export class EsploraProvider
                                 batch,
                             );
 
-                            for (const [k, v] of saved) {
+                            for (const [k, v] of result.pendingOutputs) {
                                 pendingOutputs.set(k, v);
                             }
+                            if (result.txData)
+                                allBlockTxData.push(result.txData);
                         }, this),
                     );
 
@@ -203,6 +209,14 @@ export class EsploraProvider
                         spentOutpoints,
                         pendingOutputs,
                     );
+
+                    if (isLastBatch) {
+                        this.storageService.saveSilentBlock(
+                            batch,
+                            height,
+                            encodeSilentBlock(allBlockTxData),
+                        );
+                    }
 
                     state.indexedBlockHeight = height;
                     state.lastProcessedTxIndex = i + this.batchSize - 1;
--- a/src/block-data-providers/esplora/provider.ts
+++ b/src/block-data-providers/esplora/provider.ts
@@ -128,8 +128,9 @@ export class EsploraProvider
             let height =
                 ((await this.traceReorg()) ?? state.indexedBlockHeight) + 1;
 
-            let nextBlockHash: Promise<string> | null =
-                this.getBlockHash(height);
+            let nextBlockHash: Promise<string> | null = this.prefetch(
+                this.getBlockHash(height),
+            );
 
             for (height; height <= tipHeight; height++) {
                 const blockHash = await nextBlockHash;
@@ -139,7 +140,7 @@ export class EsploraProvider
 
                 nextBlockHash =
                     height + 1 <= tipHeight
-                        ? this.getBlockHash(height + 1)
+                        ? this.prefetch(this.getBlockHash(height + 1))
                         : null;
 
                 await this.processBlock(height, blockHash);
@@ -152,7 +153,6 @@ export class EsploraProvider
     private async processBlock(height: number, hash: string) {
         const state = await this.getState();
         const txids = await this.getTxidsForBlock(hash);
-        const allBlockTxData: TransactionData[] = [];
 
         for (
             let i = state.lastProcessedTxIndex + 1;
@@ -207,8 +207,6 @@ export class EsploraProvider
                             for (const [k, v] of result.pendingOutputs) {
                                 pendingOutputs.set(k, v);
                             }
-                            if (result.txData)
-                                allBlockTxData.push(result.txData);
                         }, this),
                     );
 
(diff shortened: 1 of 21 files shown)


src/block-data-providers/esplora/provider.ts
```diff
@@ -215,15 +229,33 @@ export class EsploraProvider
...
+                batch,
+                height,
+                encodeSilentBlock(blockTxData),
+            );
+        });
+
+        // Emitted once per block, after the blob is stored: a per-batch emit
+        // broadcast the same height repeatedly, each time re-encoding whatever
+        // subset of the block had been committed so far.
+        this.eventEmitter.emit(INDEXED_BLOCK_EVENT, height);
```
chaitika: should this readback also be wrapped in a try-catch block like the per-batch loop above?

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -1,78 +1,154 @@
...
+        const repairState = await this.storageService.getOperationState(
+            SILENT_BLOCK_REPAIR_STATE,
+        );
+        const repair =
+            (repairState?.state?.version ?? 0) < SILENT_BLOCK_REPAIR_VERSION;
+
+        let written = 0;
+
+        // One height at a time: read, encode and commit run with no awaited
+        // yield between them, so a concurrent indexer or reorg write can't be
```
chaitika: the comment here says read, encode and commit happens with "no awaited yield" in between, but that looks stale — there's an await at the transaction readback and  another at dbTransactionService.execute()'s commit

src/silent-blocks/silent-blocks.gateway.ts
```diff
@@ -16,21 +35,149 @@ export class SilentBlocksGateway
...
+    }
+
+    @SubscribeMessage('ping')
+    handlePing(@ConnectedSocket() client: WebSocket): void {
+        this.sendControl(client, 'pong', null);
+    }
+
     broadcastSilentBlock(silentBlock: Buffer) {
         for (const client of this.server.clients) {
+            if (this.syncing.has(client)) continue;
```
chaitika, line 178: `broadcastSilentBlock` skips any client in syncing, and `handleSync` fixes `to` at sync start. A block indexed between `sync-start` and `sync-end` is excluded from both the stream (past the fixed `to`) and the broadcast (client is in syncing) — it's dropped with no signal to the client. 

Should broadcasts received during a sync be queued and flushed after, or should `handleSync` re-check the tip and do a follow-up sync for anything indexed during the window before acking synced?

src/silent-blocks/silent-blocks.service.ts
```diff
@@ -85,7 +161,89 @@ export class SilentBlocksService {
                 filterSpent,
             );
 
-        return this.encodeSilentBlock(transactions);
+        return encodeSilentBlock(transactions);
+    }
```
chaitika, line 167: `/range` and `WS handleSync` both clamp to the indexed tip so an un-indexed height isn't presented as an empty block. `getSilentBlockByHeight/getSilentBlockByHash` don't — a height past the tip returns the same 2-byte empty encoding as a genuinely empty, confirmed block. Should this get the same tip check?

src/block-data-providers/esplora/provider.ts
chaitika, line 274: For a coinbase-only block, the batch loop above (starting around line 157) never runs — i starts at 1, txids.length === 1 — so indexedBlockHeight/BlockState never get persisted for this height. But this emit, and the silent-block save above it, run unconditionally. A crash before the next block commits means restart reprocesses this height and re-emits INDEXED_BLOCK_EVENT for a block already broadcast. Should the state/BlockState commit happen for every block regardless of tx count?

notTanveer: yeah, wrapped the whole `processBlock` body instead. the inner catch only added a log line, so it's one catch now and the readback + blob commit are covered too.

message is `Error processing block at height X, hash Y`.

src/silent-blocks/silent-blocks.service.ts
notTanveer: that comment was stale, both the tx read and `execute()` are awaited.

rewrote it to say what the per-height loop actually buys. the old version encoded 100 heights, yielded, then wrote them all, so a whole batch of encodings could go stale. read -> encode -> commit per height keeps that window to one height, and if the indexer rewrites a height under us its own `processBlock` re-saves it. so it self-heals.

src/silent-blocks/silent-blocks.gateway.ts
notTanveer: real bug. went with the second option, re-checking the tip. queueing needs per-client buffering plus ordering against the stream, which is a lot more moving parts for the same outcome.

`handleSync` now loops: stream `[cursor, to]`, re-read the tip, stream the delta, repeat.

two things that make it hole-free and not just narrower:

- nothing awaits between the last tip read and `syncing.delete()` in the `finally`. the `nextTo` compare, the `synced` send and the delete are all synchronous, so no broadcast can land in that gap. and any block whose broadcast we skip is already in the tip we just read, since both providers commit `BlockState` in the same batch as the silent block and emit after.
- `to` only moves when another pass will actually run, so the ack never names a height we didn't stream.

capped at `MAX_CATCHUP_ROUNDS = 3` so a client syncing during IBD can't pin the socket. hitting the cap acks with `tip > to`, which is the signal to sync again.

two tests added: tip moving mid-stream, and a tip that never settles.

src/silent-blocks/silent-blocks.service.ts
notTanveer: fixed for height, not for hash.

`height/:height` now 404s when the height is above the tip. put the check in the controller and not the service. that endpoint is the only caller that can get an arbitrary height (`handleBlockIndexedEvent` passes one that was just committed), and it already reads the tip for the cache decision, so it's a reorder + a guard with no extra lookup.

hash can't get the same check. a hash carries no ordering and there is no block-hash -> height index. `BlockState` is keyed by height with the hash as the value, and `hashIndexRange` only covers transactions, so a coinbase-only block and an unknown hash both scan to zero rows. telling them apart needs a hash -> height index, which is a schema change and not a guard, so leaving it out of this PR. happy to add it if you want it. same reason `hash/:hash` is already on `max-age=60`.

src/block-data-providers/esplora/provider.ts
notTanveer: confirmed. for `txids.length === 1` the loop never runs, so `indexedBlockHeight` / `BlockState` were never persisted while the blob save and the emit ran anyway.

moved the state commit into the same batch as `saveSilentBlock`, unconditionally: `indexedBlockHeight = height`, `lastProcessedTxIndex = 0`. the batch loop still writes `lastProcessedTxIndex` for mid-block resume, but the block-done marker and the blob are one atomic write now, so they can't diverge. bitcoin core provider already does it this way.

test added in `esplora/provider.spec.ts`. fails on the old code (`saveBlockState` never called), passes now.

PR #111, opened 2026-09-20
perf: cut per-block indexing time by eliminating doomed LMDB probes

theanmolsharma:
## Why

Production indexing was taking 63-113s per block. With ~108k blocks left to catch up, that is roughly 112 days. The question on the table was whether to rewrite the indexer in Rust.

The per-block telemetry from #110 answers it:

```
block=861864 tx=3836 in=7107 out=10504 |
  processBlock=2655ms  index=212ms  markSpent=81102ms  commit=28943ms  total=112983ms
```

`processBlock` (RPC + JSON parse) and `index` (secp256k1 ECDH, tagged hashes, tweak derivation) are the only CPU-bound phases a language change touches. Together they are **2.9s of 113s — 2.6%**. An infinitely fast rewrite of that work takes the block from 113s to 110s. The other 97.4% is LMDB, which is the same C library under Rust bindings. So: not Rust. `markSpent` (72%) and `commit` (26%) are the target, and both are fixable here.

## Root cause

`markOutputsSpent` probed **every input of every transaction** — 7,107 on this block. But the indexer only ever writes an `out:` record for P2TR outputs of BIP-352-eligible transactions, so an input spending anything else is a guaranteed miss, and an LMDB miss still pays a full B+tree descent before returning null.

## What changed

- **Skip outpoints that cannot be in the index.** Filter on the previous output's script before queueing. A superset filter with no false negatives — the skipped probes could only ever have returned null, so behaviour is unchanged. `prevOutScript` is already in memory for free (Core >= 23 means verbosity 3, which embeds `prevout.scriptPubKey`), so no extra RPC.
- **Sort probes by encoded key.** `out:` keys are raw-txid-prefixed, so spend order is random across the B+tree; sorting keeps shared upper-level pages hot.
- **Retire the `idx:us:` unspent index.** It cost a put per output created and a del per spend (~1/3 of all write ops) and made reads *slower*: the filtered path scanned the index then re-read each record, where one `out:<txid>` scan filtered on the `isSpent` byte does it in a single pass. That byte is already authoritative, so **no reindex is needed** — pre-existing `idx:us:` keys are inert.
- **Telemetry** gains `probes`/`hits`/`skipped`, so the remaining cost can be attributed.

Plus three bugs found on the way, each its own commit:

- **RPC password logged in cleartext.** The retry path stringified the whole axios config, including `auth`, on every retry. Redacted — but **any password running with a previous build should be rotated.**
- **`isSyncing` could wedge permanently.** The flag was set before the `try/finally` that clears it, with `getState()` in the unguarded gap. If it threw, the indexer stopped silently until restart.
- **Stale RPC sockets.** Explicit keep-alive agent with a 20s idle timeout, under Core's 30s `rpcservertimeout` — the EPIPE/ECONNRESET between blocks. Also stopped eagerly pretty-printing multi-MB verbosity-3 `getblock` payloads into a verbose log that then discards them.

## Testing

- 89 unit tests pass (16 new). New `storage.service.spec.ts` covers same-block spends via `pendingOutputs`, unindexed outpoints, repeat spends, order independence, and the `filterSpent` read path — the correctness claim behind the prefilter.
- All 5 e2e tests pass, including the real P2TR spend case.
- E2E was failing before any of this, on `main` too: `waitForBitcoind` polled inside a try/catch, but the helper's `request()` swallows axios errors and resolves with an error envelope rather than rejecting, so the wait returned on the first attempt before bitcoind was listening. Fixed in a separate commit.

## Expected effect, and one thing that still doesn't add up

Projected, not yet measured against production: `markSpent` ~81s -> ~12s, `commit` down about a third.

At 16GB + NVMe, 11.4ms per lookup is still ~100x worse than expected. Every change here reduces lookup *count*, which helps regardless of the reason — but the new `probes`/`hits` counters are what will say whether the remainder is volume or per-lookup latency. **Worth reading that line before reaching for the next lever.** Deferred, largest first: multi-block commits during catch-up (the big one at this block count), an in-memory Bloom filter over stored outpoints, block prefetch, and re-evaluating `compression: true`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


theanmolsharma:
## Follow-up: `commit` was the remaining 80%, now batched

Production numbers after the prefilter landed (blocks 862597-862599):

```
markSpent=1520ms (probes=282 hits=266 skipped=7485)  commit=22441ms  total=26829ms
markSpent=2761ms (probes=332 hits=307 skipped=7161)  commit=33999ms  total=40070ms
markSpent=6619ms (probes=1033 hits=907 skipped=5623) commit=21017ms  total=31970ms
```

The prefilter did its job — `markSpent` went 81,102ms -> ~3,600ms mean (~22x), with `skipped` confirming ~96% of probes were never worth issuing. Mean block: 82s -> 33s. But that just promoted `commit` to ~80% of every block.

### It is not the commit code path

20,000 puts commit in **16ms** against a small local DB. The problem is random insertion into a *large* B+tree: `out:` keys are prefixed by raw txid, so writes scatter, and copy-on-write rewrites the root-to-leaf path for every touched leaf. Same 15,000 ops, same code, growing DB:

| keys in DB | commit for 15k ops |
|---|---|
| 0.3M | 41ms |
| 0.9M | 149ms |
| 1.8M | 245ms |

Committing per block pays for those shared interior pages every single block.

### Batching them

Measured over identical total work:

| blocks/commit | per block | vs per-block |
|---|---|---|
| 1 | 113.0ms | — |
| 5 | 30.1ms | 3.76x |
| **25** | **21.4ms** | **5.28x** |
| 50 | 19.5ms | 5.79x |

Defaults to 25 (`db.commitBatchBlocks`). Catch-up only — at the tip there is one block to write, so the batch is a single block and visibility is unchanged.

Two invariants this had to respect, both easy to get wrong:
- `pendingOutputs` now spans the batch. An output created in one block and spent in a later block of the same batch isn't committed yet, so the spend resolves only from that map.
- Block state is still written **per height**. `traceReorg` walks it one height at a time; collapsing it would break reorg detection and leave intermediate blocks uncleaned.

### Also ruled out, by measurement rather than reasoning

- **`compression: true`** — no effect (547MB vs 546MB, same commit time). Dropping it was on my deferred list; it's off now.
- **`NO_SYNC_FLUSH`** — our `transactionSync()` passes no flags, which masks to `0`, so LMDB fsyncs on every commit while lmdb-js's own paths pass `MDB_NOSYNC` under `overlappingSync`. Looked like a certain win from the source. Benchmarked at **1.00x**. Not pursuing it.

### Verification

e2e ran with batching on and the log confirms the real path: `blocks=1-25 (25)` through catch-up, degrading to `blocks=201-207 (7)` near the tip. All 5 e2e and 89 unit tests pass.

### Expected

commit ~26s -> ~5s/block, so ~33s -> ~12s. Combined with the prefilter that's 82s -> ~12s (~7x), or roughly 112 days -> 15 days of catch-up. Projected from local measurements; the production numbers will tell.

Still worth a look once this is running: reads are ~6ms/probe with a ~90% hit rate, which is page-cache misses against a multi-GB DB. If that holds after batching, more RAM is the lever, not more code.

PR #112, opened 2026-09-20
perf: batch blocks into one write transaction while catching up

theanmolsharma:
Follow-up to #111. That PR removed the doomed LMDB probes; this one goes after what it promoted to the bottleneck.

## Where things stand after #111

Production, blocks 862597-862599:

```
markSpent=1520ms (probes=282 hits=266 skipped=7485)  commit=22441ms  total=26829ms
markSpent=2761ms (probes=332 hits=307 skipped=7161)  commit=33999ms  total=40070ms
markSpent=6619ms (probes=1033 hits=907 skipped=5623) commit=21017ms  total=31970ms
```

The prefilter worked: `markSpent` went 81,102ms -> ~3,600ms mean (~22x), and `skipped` confirms ~96% of probes were never worth issuing. Mean block 82s -> 33s. But `commit` is untouched at ~26s, so it is now **~80% of every block**.

## It is not the commit code path

20,000 puts commit in **16ms** against a small local database. The cost is random insertion into a *large* B+tree: `out:` keys are prefixed by raw txid, so writes scatter across the tree, and LMDB's copy-on-write rewrites the path from the root down to each touched leaf on every commit. Same 15,000 ops, same code, growing database:

| keys in DB | commit for 15k ops |
|---|---|
| 0.3M | 41ms |
| 0.9M | 149ms |
| 1.8M | 245ms |

Committing per block pays for those shared interior pages every single block. That is what this changes.

## Batching

Measured over identical total work:

| blocks/commit | per block | vs per-block |
|---|---|---|
| 1 | 113.0ms | — |
| 5 | 30.1ms | 3.76x |
| **25** | **21.4ms** | **5.28x** |
| 50 | 19.5ms | 5.79x |

Defaults to 25 via `db.commitBatchBlocks`. Returns flatten past that, while larger batches hold more in memory and re-index more after a crash.

**Catch-up only.** At the tip there is one block to write, so the batch is naturally a single block and visibility is unchanged.

## Two invariants this had to respect

Both are easy to get wrong and neither is caught by types:

- **`pendingOutputs` now spans the batch**, not one block. An output created in one block and spent in a later block of the same batch is not committed yet, so the spend can only be resolved from that map. Scoping it per block would silently leave those outputs marked unspent.
- **Block state is still written per height**, not once per batch. `traceReorg` walks it one height at a time; collapsing it would break reorg detection and leave intermediate blocks' transactions uncleaned on a reorg.

Block-indexed events are emitted after the commit — a listener that reads back before then would see uncommitted state.

## Ruled out by measurement

Both of these were on my deferred list and looked plausible from the source. Neither survived a benchmark:

- **`compression: true`** — no effect. 547MB vs 546MB, same commit time.
- **`NO_SYNC_FLUSH`** — our `transactionSync()` passes no flags, which masks to `0`, so LMDB fsyncs on every commit, while lmdb-js's own write paths pass `MDB_NOSYNC` under `overlappingSync`. It reads like a certain win. Benchmarked at **1.00x**.

## Verification

89 unit tests and all 5 e2e tests pass. e2e ran with batching enabled and the indexer log confirms the real path was exercised, including the degradation near the tip:

```
blocks=1-25 (25) ... blocks=176-200 (25)
blocks=201-207 (7)  blocks=208-214 (7)  blocks=229-236 (8)
```

Telemetry now reports a height range and per-block cost.

## Expected effect

commit ~26s -> ~5s/block, so ~33s -> ~12s. Combined with #111 that is 82s -> ~12s (~7x), roughly 112 days -> 15 days of catch-up. Projected from local measurements — production will be the real test.

One thing still worth watching: reads are ~6ms/probe at a ~90% hit rate, which is page-cache misses against a multi-GB database. If that persists after this lands, more RAM is the lever, not more code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


theanmolsharma:
## Production results — default lowered to 5

Ran against three batch sizes. Commit time **per block**:

| blocks/commit | commit/block | markSpent/block | total/block |
|---|---|---|---|
| 1 | 25.8s | 3.6s | 33.0s |
| **5** | **14.6s** | **4.9s** | **21.4s** |
| 25 | 14.3s | 6.2s | 21.8s |

Amortisation is real from 1 to 5, then flat. 25 buys ~2% on commit while costing 5x the memory, 5x the crash re-index window, and 5x the visibility delay — and it measurably *hurt* reads (markSpent 4.9s -> 6.2s per block), consistent with the larger pending-write set evicting LMDB pages from page cache.

So the 25 default in the first commit was wrong, and instructively so: it came from a local benchmark on a database that still fit in page cache, which is exactly the regime the production box is not in. Defaulted to 5 and recorded the reasoning in the constant so it doesn't get re-derived from the same misleading measurement.

Net across #111 and this PR: **82s -> 21.4s per block (~3.8x)**, catch-up ~99 days -> ~26 days.

## This is the end of the code levers

Commit is still 68% of every block at ~2.5ms per database operation. LMDB does 16µs/op in RAM — a ~150x gap that is page-fault latency, not code.

Six optimisations measured and rejected along the way, which I'd rather record than have someone retry:

| idea | result |
|---|---|
| Rewrite in Rust | 1.03x ceiling (CPU is 2.6% of block time) |
| `compression: true` off | 1.00x (547MB vs 546MB) |
| `NO_SYNC_FLUSH` on `transactionSync` | 1.00x |
| Sort write ops by key before commit | **0.66x — 1.5x slower** |
| Batch >5 blocks per commit | ~1.0x, and worse reads |
| Dedupe the redundant per-tx time-index write | 1.09x in RAM, ~0 in production |

That last one is a genuine redundancy — `encodeTimeIndexKey` contains only block-level data, so every eligible tx in a block rewrites the identical key, ~19% of all write ops. But duplicate writes land on an already-dirty cached page, so they cost far less than unique ones: removing 19% of ops saved only 8.8% of commit time in RAM, and would save less in production where the unique ops are the ones paying the 2.5ms. Left alone rather than shipped as a fake win.

What remains is the machine: the working set no longer fits in RAM, and each resulting fault is costing milliseconds rather than the ~0.1ms local NVMe would.

